PR #17, opened 2022-11-02
Feature/GitHub actions base

lakooi:


.github/workflows/mainpush.yaml
```diff
@@ -0,0 +1,13 @@
+name: PR Workflow
+run-name: ${{ github.actor }} pull request
+on:
+  push:
+    branches:
+      - main
+jobs:
+  TestMain:
```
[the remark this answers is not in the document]
lakooi: Lets dicuss how we want to format files in the project after today's daily. I agree with your remarks.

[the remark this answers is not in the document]
no-comma: We did decide to use `camelCase` for JavaScript / TypeScript code, but for these files it may still be better to use snake or kebab case because all of the default syntax seems to be in those styles.

.github/workflows/pr-workflow.yml
```diff
@@ -0,0 +1,48 @@
...
+    defaults:
+      run:
+        working-directory: ./client
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: '18'
+      - name: Install dependencies
+        run: npm install
```
lakooi, line 26: Should this be npm ci or npm install?

.github/workflows/pr-workflow.yml
```diff
@@ -0,0 +1,48 @@
+name: PR Workflow
+run-name: ${{ github.actor }} pull request
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    types:
+      - opened
```
lakooi, line 9: The commits done to the feature branch haven't triggered new runs of github actions. I suppose it is because it runs only for types: opened.

I suppose the type -edited could be added here as well:

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request

.github/workflows/pr-workflow.yml
no-comma: It can be changed to `npm ci` in the PR which pushes `package-lock`.

lakooi:
I deleted the types part of the pull request, as adding the -edited did not trigger the actions. They were now triggered

client/.eslintrc.json
```diff
@@ -0,0 +1,46 @@
+{
+    "env": {
+        "browser": true,
+        "es2021": true,
+        "node": true
```
Gskartwii, line 5: I'm not sure I understand `"node": true` here. The client code should run only in the browser, right?

client/.eslintrc.json
```diff
@@ -0,0 +1,46 @@
+{
+    "env": {
```
Gskartwii, line 2: Should we also have an indent width of 2 in this file?

server/.eslintrc.json
```diff
@@ -0,0 +1,38 @@
+{
+    "env": {
```
Gskartwii, line 2: See comment about indent width in `client/.eslintrc.json`

Gskartwii (changes requested):
This PR is invaluable, thanks! I've asked for some clarifications wrt the eslintrcs

client/.eslintrc.json
no-comma: It was automatically generated this way, but I can change it for the sake of consistency.

client/.eslintrc.json
no-comma: I wasn't sure whether it's necessary since the client code should run in the browser, but Node does need to be installed to run the frontend. From a quick search now it doesn't seem to be necessary, I'll remove it.

client/.eslintrc.json
lakooi: In a way I feel that this is a decision to be made by the whole team ( Amount of indentation ), but we can look into it during a retro if it is deemed important later on. Lets just go with the 2 this time.

jyrge (approved):

Issue #9, opened 2022-10-26
Development environment and tools documentation


no-comma:
How to set up your environment. (Probably a README)
Different READMEs for frontend and backend?
Downloading linters?

Issue #8, opened 2022-10-26
Product base setup


no-comma:
After tech stack has been selected, create a base that displays something like "Hello world" when run.

Depends on #1 
Depends on #2

no-comma:
What kind of a directory structure should we use? `/client` for frontend and `/server` for backend, each with necessary subdirectories?

viivi-uhari:
@no-comma 's proposal for the directory structure seems good! We could discuss how to link the frontend to the backend on Monday after the daily. At this point running the application locally and using different ports for frontend and backend could be a simple and working approach.

viivi-uhari:
Pull requests made for frontend #19 and backend #18 bases.

PR #18, opened 2022-11-06
Backend base

no-comma:
Adds a backend base with a hello world program and basic instructions for running it. Windows & macOS users especially should verify that the instructions are correct and expand on them.

Related issues: #8 #9

### DIFF
--- a/server/index.ts
+++ b/server/index.ts
@@ -1,0 +1,12 @@
+import express, { Application, Request, Response } from 'express';
+
+const app: Application = express();
+const port: number = 3000;
+
+app.get('*', (req: Request, res: Response) => {
+  res.send(`Hello ${req.path}`);
+});
+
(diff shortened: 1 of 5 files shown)


server/index.ts
```diff
@@ -0,0 +1,12 @@
+import express, { Application, Request, Response } from 'express';
+
+const app: Application = express();
+const port: number = 3000;
```
lakooi, line 4: I think this is good for the base for now, but at some point port number is preferably passed through an environmental variable.

no-comma: @Gskartwii has already implemented this at https://github.com/Gskartwii/aalto-grades-base/tree/backend-base

lakooi (approved):

jyrge (approved):

Issue #28, opened 2022-11-09
Design and implement login and signup page (Frontend)


viivi-uhari:
Login and sign up (to specify whether the user is a teacher or student) pages needed. Sign up will eventually be deleted.

no-comma:
This issue is to be replaced with more specific ones.

Issue #15, opened 2022-11-01
EPIC: Basic user management


PasiSa:
Implement a way to create new user accounts and log in using existing user accounts. Simple local sign in is sufficient for now, but later on external authentication methods should be the primary login method (e.g., using Aalto IdP). Users should be associated with name, email. They also have role respective to course (teacher/student). Some users are sysadmins with full rights.

Goal: UI to sign up as a new user (with needed information), or to log in as an existing user

- [x] #28 
- [x] #36 
- [x] #55 
- [x] #57 
- [x] #58

viivi-uhari:
UI elements, API (+ user model), backend authentication (tokens). During sign up specify if student or teacher etc.

viivi-uhari:
Estimate 8

Issue #14, opened 2022-11-01
EPIC: Design initial data model


PasiSa:
Design and implement initial data model for the system. It is likely that we need to iterate and adjust the model as the system develops, but the basic concepts probably include at least the following:

Course
- Course code, name

Course instance
- ==> Course
- Instance name (e.g. fall 2022)
- Teachers (==> User)
- Start / end time

User
- Name
- Email
- Student ID
- Role (Student/Teacher/sysadmin)  -- may not be direct table member, because e.g. teacher role is respective to specific course

Grade (this needs some thought: student will have the final grade for each course, that is calculated from partial grades. Partial grades are in turn calculated from exercise/exam/etc. points. In principle there is an hierarchical structure of points/grades)
- ==> Course instance
- Expiry time

Goal: Diagram of the data model, with main attributes for each class

- [ ] #31

lakooi:
Estimate 5

viivi-uhari:
Design with frontend developers

PR #34, opened 2022-11-11
Basic user management

Gskartwii:
Implements the login and signup routes. 

Related: #15, #28


Gskartwii:
Requires #14

Gskartwii:
Dependent on #51

Gskartwii:
Need to think carefully about what kind of result the login and signup routes should return, and how sessions are securely handled. Access control should be implemented server-side.

server/src/auth.ts
```diff
@@ -0,0 +1,18 @@
+// SPDX-FileCopyrightText: 2022 The Aalto Grades Developers
+//
+// SPDX-License-Identifier: MIT
+
+export type PlainPassword = string;
+export type UserRole = 'Teacher' | 'Student' | 'Admin';
```
no-comma, line 6: Use an enum instead?

server/src/index.ts
```diff
@@ -3,11 +3,125 @@
...
+interface LoginRequest {
+  username: string,
+  password: PlainPassword,
+}
+
+interface SignupRequest {
+    username: string,
+    password: PlainPassword,
+    email: string,
+    role: UserRole,
```
no-comma, line 22: Why are there 4 spaces here? Does this really pass the linter?

server/src/index.ts
```diff
@@ -3,11 +3,125 @@
...
+
+const validateSignupFormat = (body: any): body is SignupRequest =>
+  body &&
+    body.username &&
+    body.password &&
+    body.email &&
+    validateUserRole(body.role) &&
+    typeof body.username === 'string' &&
+    typeof body.password === 'string' &&
+    typeof body.email === 'string';
```
Tonits, line 46: We have decided to use the function keyword for named functions, and arrow functions for anonymous functions. These functions should be changed so that they use the function keyword.

Gskartwii: Fixed up!

server/src/auth.ts
Gskartwii: Yes, that's much better.

Gskartwii:
In the latest merge commit, I took in changes in `client/src/App.js` from both `main` and `feature/auth` and tried to merge them together. I wasn't fully confident in what I did, so someone might want to check that!

client/src/components/auth/Login.js
```diff
@@ -0,0 +1,44 @@
...
+//
+// SPDX-License-Identifier: MIT
+
+import React, {useState} from 'react';
+import userService from '../../services/user';
+import { useNavigate } from 'react-router-dom';
+import LoginForm from './LoginForm';
+import { useTheme } from '@mui/material/styles';
+
+const Login= () => {
```
no-comma, line 11: Missing space? `const Login = () => {`

server/src/index.ts
```diff
@@ -3,17 +3,81 @@
...
+  password: PlainPassword,
+  email: string,
+  role: UserRole,
+}
+
+function validateUserRole(role: any): role is UserRole {
+  return typeof role === 'string' && (
+    role === 'Teacher' ||
+    role === 'Student' ||
+    role === 'Admin'
```
no-comma, line 30: If these validate functions are to stay, this is seems outdated with changing UserRole to enum.

server/src/index.ts
```diff
@@ -3,17 +3,81 @@
...
+
+function validateLoginFormat(body: any): body is LoginRequest {
+  return body &&
+    body.username &&
+    body.password &&
+    typeof body.username === 'string' &&
+    typeof body.password === 'string';
+}
+
+function validateSignupFormat(body: any): body is SignupRequest {
```
no-comma, line 42: I wonder if these validate functions are really necessary, TypeScript may have some better way to achieve the same thing, but I'm not sure.

no-comma (commented):
There is some inconsistency in spacing inside brackets in frontend imports. Sometimes imports are written as `import { foo } from 'bar';` and others `import {foo} from 'bar';`. ESLint can probably be configured to enforce one or the other.

server/src/index.ts
```diff
@@ -3,17 +3,81 @@
...
+  return body &&
+    body.username &&
+    body.password &&
+    body.email &&
+    validateUserRole(body.role) &&
+    typeof body.username === 'string' &&
+    typeof body.password === 'string' &&
+    typeof body.email === 'string';
```
no-comma, line 50: Why are these type checks here? These seem like things that would already be ensured by `body is SignupRequest`.

Gskartwii: `body is SignupRequest` is a way of indicating that we have checked that `body` is of the correct type/interface. It's basically a hint to the type checking engine, and it doesn't have a runtime effect.

server/src/index.ts
```diff
@@ -30,6 +94,55 @@ app.get('/v1/user/courses', async (req: Request, res: Response) => {
...
+    // carry this info
+    res.status(400);
+    return res.send({
+      success: false,
+      error: error,
+    });
+  }
+});
+
+app.post('/v1/auth/signup', express.json(), (req: Request, res: Response) => {
```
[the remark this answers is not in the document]
Gskartwii: You're right! I should remove this ASAP

server/src/index.ts
no-comma: As far as I can tell as well it doesn't seem like TypeScript itself offers any nice way to do this. However, in #116 I used [the yup library](https://github.com/jquense/yup) for validation, which in my opinion is nicer solution.

client/src/hooks/useLogout.js
```diff
@@ -0,0 +1,30 @@
...
+
+// custom hook for logout, used since no individual logout page will be created
+
+const useLogout = () => {
+
+  const { setAuth } = useAuth();
+  
+  const logout = async () => {
+
+    setAuth({});
```
marillam, line 16: Can this be set before the logout() below has been successful? Does this code enable a situation where the logout fails (so the user is still "logged in") but the auth (context) is empty.

server/src/index.ts
```diff
@@ -0,0 +1,28 @@
+// SPDX-FileCopyrightText: 2022 The Aalto Grades Developers
+//
+// SPDX-License-Identifier: MIT
+
+/*import { port } from './configs';
```
marillam, line 5: If not needed, delete?

marillam (changes requested):
Not necessarily required changes, but I had some questions and ideas. Also as a general comment: I noticed that the server code doesn't use arrow functions. Not saying they should be used, but is there a particular reason for this.

no-comma (changes requested):
Great that this is finally ready for review!

One overarching issue is in regards to whether the email or username is used for login. There seems to be a bit of incoherency in regards to this, for example, the client login page has a field for entering a username, but in the database the entered value is matched with the email.

Another thing to note is that our user table currently does not have an attribute for a username, only the real name of the user and their email.

docker-tests/docker-compose.yaml
```diff
@@ -21,6 +21,7 @@ services:
       POSTGRES_USER: "postgres"
       POSTGRES_DATABASE: "postgres"
       POSTGRES_PASSWORD: "${POSTGRES_PASSWORD}"
+      AALTO_GRADES_TEST_ENVIRONMENT: "true"
```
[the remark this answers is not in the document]
Gskartwii: The intent of this is to control the behavior of what happens when running unit tests.

In more detail, the Supertest framework isn't compatible with the `Secure` flag on cookies, so we need to disable it only in that environment. For localhost development, that flag is useful to have.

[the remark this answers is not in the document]
no-comma: If this is something that only concerns unit tests then it has significant overlap with `SERVER_RUN_TESTS` below and the two should probably be merged.

server/src/controllers/auth.ts
```diff
@@ -0,0 +1,83 @@
...
+    }
+  });
+
+  if (exists !== null) {
+    throw new UserExists();
+  }
+
+  try {
+    const model: User = await User.create({
+      name: username,
```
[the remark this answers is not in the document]
Gskartwii: Thanks for the clarification, I'll update all code accordingly!

server/.eslintrc.json
```diff
@@ -34,6 +34,7 @@
       "error",
       "always"
     ],
+    "@typescript-eslint/no-explicit-any": "off",
```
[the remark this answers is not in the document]
Gskartwii: I've removed the explicit `any`, replacing many of them with `unknown`. What are your thoughts on this solution?

server/src/controllers/auth.ts
```diff
@@ -0,0 +1,83 @@
...
+  constructor() {
+    super('credential format is invalid, possibly bad email');
+  }
+}
+
+export async function validateLogin(username: string, password: PlainPassword): Promise<LoginResult> {
+  const user: User | null = await User.findOne({
+    attributes: ['id', 'password'],
+    where: {
+      email: username,
```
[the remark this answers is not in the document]
Gskartwii: Good catch!

server/test/routes/auth.test.ts
```diff
@@ -0,0 +1,101 @@
...
+//
+// SPDX-License-Identifier: MIT
+
+import { app } from '../../src/app';
+import supertest, { SuperAgentTest } from 'supertest';
+import { UserRole } from '../../src/controllers/auth';
+
+const request: supertest.SuperTest<supertest.Test> = supertest(app);
+
+describe('Test login route', () => {
```
no-comma, line 11: In our other tests we've used descriptions in the format of `Test METHOD /v1/some/route`, the test descriptions here should follow the same convention.

server/src/controllers/auth.ts
```diff
@@ -0,0 +1,80 @@
...
+// SPDX-License-Identifier: MIT
+
+import User from '../database/models/user';
+import argon from 'argon2';
+
+export type PlainPassword = string;
+export enum UserRole {
+  Student = 'STUDENT',
+  Teacher = 'TEACHER',
+  Admin = 'ADMIN',
```
no-comma, line 12: The database enum in `server/src/database/models/courseRole.ts` calls this `'SYSADMIN'`, one of the two should be changed to be consistent.

server/src/configs/index.ts
```diff
@@ -0,0 +1,19 @@
+// SPDX-FileCopyrightText: 2022 The Aalto Grades Developers
+//
+// SPDX-License-Identifier: MIT
+
+const parsedPort: number = Number(process.env.AALTO_GRADES_BACKEND_PORT);
+const port: number = isNaN(parsedPort) ? 3000 : parsedPort;
+const testEnv: boolean = process.env.AALTO_GRADES_TEST_ENVIRONMENT == 'true';
+const jwtSecret: string =  process.env.AALTO_GRADES_JWT_SECRET || 'TOP_SECRET';
+const frontendOrigin: string = process.env.AALTO_GRADES_FRONTED_CORS_ORIGIN || 'http://localhost:3005';
+if (jwtSecret == 'TOP_SECRET') {
```
no-comma, line 10: I think an empty line before this if statement would look cleaner.

client/src/components/auth/Login.js
```diff
@@ -0,0 +1,45 @@
...
+
+  const navigate = useNavigate();
+  const [errorMessage, setErrorMessage] = useState('');
+  const theme = useTheme();
+
+  const loginUser = async (userObject) => {
+    try {
+      const response = await userService.login(userObject);
+      
+      // if login is successfull, save user role to context
```
no-comma, line 24: Successful

server/src/routes/login.ts
```diff
@@ -0,0 +1,183 @@
...
+  studentID: string,
+  role: UserRole,
+}
+
+const signupSchema: yup.AnyObjectSchema = yup.object().shape({
+  username: yup.string().required(),
+  password: yup.string().required(),
+  email: yup.string().required(),
+  studentID: yup.string().required(),
+  role: yup.string().oneOf(['ADMIN', 'STUDENT', 'TEACHER', 'ASSISTANT']).required(),
```
no-comma, line 28: Use the TypeScript enums here:
```
role: yup.string().oneOf([UserRole.Admin, UserRole.Student, UserRole.Teacher, UseRole.Assistant]).required(),
```

no-comma (commented):
Just more minor comments, the backend is starting to look good, but the frontend still has the issue of talking about usernames.

marillam (approved):
Good changes

docker-tests/docker-compose.yaml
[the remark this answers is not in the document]
Gskartwii: Fully agreed. I wonder what could be a good name for such a variable.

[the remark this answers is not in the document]
no-comma: I think just using `AALTO_GRADES_TEST_ENVIRONMENT` is fine enough for that. Remember to update GitHub action and Docker configurations if we choose to merge the variables.

no-comma (approved):
There are still some minor details that could be addressed, but nothing major enough to warrant delaying merging, LGTM!

PR #39, opened 2022-11-16
Apply license according to REUSE

no-comma:
Uncommentable files are marked in `.reuse/dep5` to avoid `.license` files from being created and the Expat (MIT) License is included in `LICENSE` at the root of the project anyway so GitHub can recognize and display it.


Gskartwii (approved):
Awesome!

lakooi (approved):

PR #40, opened 2022-11-23
Add SSH config with help for forwarding ports from the dev server

Gskartwii:
Note that this will require you to have configured the usernames for the corresponding hosts in your own SSH config (`~/.ssh/config`), e.g.

```
Host magi
    User hannula7
Host grades
    User hannula7
```

This port forwarding *should* only make the services available from your local machine (not even the local network beyond that), as long as you're careful with not forwarding them any further.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,3 +6,40 @@ SPDX-License-Identifier: MIT
 
 # base-repository
 Repo for Aalto-Grades Program
+
+## Development environment
+
+As one of the developers: in order to access the development environment, you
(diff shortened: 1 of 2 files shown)


lakooi (approved):

no-comma (changes requested):
Looks good, you should maybe still merge main and ensure REUSE compliance.

no-comma (approved):
LGTM!

Issue #27, opened 2022-11-09
Design backend structure


jyrge:
Design file/directory structure for backend

jyrge:
Proposing following directory structure for backend:

```
server/
    src/
        models/         // Database models
        services/
        routes/
        configs/
    test/               // Unit tests
        <similar structure as in src>
```

PR #41, opened 2022-11-23
Backend file structure

Tonits:
Added file and folder structure, with READMEs to make structure visible.
#27


Gskartwii (changes requested):
Some Git conflicts need to be resolved prior to merging. Be sure to also add REUSE headers to the READMEs (otherwise the CI won't let you through!)

lakooi (approved):

Gskartwii (approved):
Thanks!

PR #71, opened 2022-12-02
Front page base

no-comma:
All functionality is not included, but for the sake of being able to work in smaller feature branches this should be merged.
Related: #72


client/src/services/courses.js
```diff
@@ -0,0 +1,13 @@
+// SPDX-FileCopyrightText: 2022 The Aalto Grades Developers
+//
+// SPDX-License-Identifier: MIT
+
+import axios from './axios';
+
+const courses = async () => {
```
kiiamakir, line 7: Should probably be named getCourses since additional services for adding/modifying courses might be added.

client/src/components/FrontPage/BasicGrid.js
```diff
@@ -0,0 +1,31 @@
...
+// SPDX-License-Identifier: MIT
+
+import React from 'react';
+import PropTypes from 'prop-types';
+import Grid2 from '@mui/material/Unstable_Grid2';
+import CourseCard from './CourseCard';
+
+// disableEqualOverflow ?
+// is course code unique for all here?
+// xs, sm etc
```
kiiamakir, line 12: Comments here are a bit unclear.

kiiamakir (changes requested):

client/src/services/courses.js
no-comma: Changed.

client/src/components/FrontPage/BasicGrid.js
no-comma: @marillam Can you clarify?

marillam: Those were more questions than anything, currently code uses course code as unique key but should probably change when there is a unique identifier available. Rest are for grid element styling, I can delete them.

client/src/components/FrontPage/CourseTable.js
```diff
@@ -0,0 +1,47 @@
...
+import React from 'react';
+import PropTypes from 'prop-types';
+import Table from '@mui/material/Table';
+import TableBody from '@mui/material/TableBody';
+import TableHead from '@mui/material/TableHead';
+import CourseTableRow from './CourseTableRow';
+import CourseHeadTableRow from './CourseHeadTableRow';
+
+//used to sort the table rows by course code
+function sortByCode(a, b) {
```
marillam, line 14: Is sortByCode generic enough to be used with other components? If so, could be moved somewhere as a service maybe.

client/src/components/FrontPage.js
```diff
@@ -0,0 +1,40 @@
...
+
+  useEffect(() => {
+    coursesService.getCourses()
+      .then((data) => setCurrentCourses(data.courses.current));
+  }, []);
+
+  useEffect(() => {
+    coursesService.getCourses()
+      .then((data) => setPreviousCourses(data.courses.previous));
+  }, []);
```
marillam, line 24: Should both courseService.getCourses() include a catch() for errors?

client/src/services/courses.js
```diff
@@ -0,0 +1,13 @@
...
+//
+// SPDX-License-Identifier: MIT
+
+import axios from './axios';
+
+const getCourses = async () => {
+  const response = await axios.get('/v1/user/courses');
+  console.log(response.data);
+  return response.data;
+};
```
marillam, line 11: Should there be a try catch here too?

marillam (commented):
Looks good, let's check error handling.

client/src/services/courses.js
kiiamakir: I think that try catch is not required here if it's included in the FrontPage component. If the get is not successful, it will return 400 or some other error which will be handled in the try catch within the component.

viivi-uhari: Okay, working on these and adding catch() to useEffect().

client/src/components/FrontPage/CourseTable.js
viivi-uhari: Okay, working on this.

marillam (approved):
Everything looks good now, changes made.

kiiamakir (approved):

PR #75, opened 2022-12-06
Modify server file structure and backend ESLint

### LINKED ISSUE #60: Modify server file structure to fit decided structure

### LINKED ISSUE #74: TypeScript config to require defining types

no-comma:
Closes #60 
Closes #74


server/src/routes/controllers/user.ts
```diff
@@ -0,0 +1,23 @@
+// SPDX-FileCopyrightText: 2022 The Aalto Grades Developers
+//
+// SPDX-License-Identifier: MIT
+
+import { Request, Response } from 'express';
+import { TeacherCourses, getTeacherCourses } from '../../services/user';
+
```
Tonits, line 7: I think it is common practice to have the controllers located at src/controllers instead of src/routes/controllers

no-comma: In that case `src/routes/index.ts` should probably also be moved to `src/routes.ts` and the `src/routes` directory can be deleted completely. I think that would be better too but we'll have to discuss that first.

Tonits (approved):

jyrge (approved):

PR #78, opened 2022-12-10
Retrieve teachers' courses from database

### LINKED ISSUE #59: Route and database connection for fetching list of courses

no-comma:
Closes #59


server/src/controllers/course.ts
```diff
@@ -17,3 +17,9 @@ export interface Course {
   name: LocalizedString,
   evaluationInformation: LocalizedString
 }
+
+export enum Language {
```
jj-stigell, line 21: Should we use enums 'EN', 'FI' and 'SV' to match the database language enums?

server/src/controllers/user.ts
```diff
@@ -3,99 +3,108 @@
...
+        return false;
+      }
+      courseFlags.add(courseInstance.courseId);
+      return true;
+    });
+
+    // Construct CourseData objects and determine whether the course is current
+    // or previous.
+    const currentDate: Date = new Date();
+    for (const i in instances) {
```
jj-stigell, line 54: I would avoid querying database inside for loop. Could affect performance negatively when instance count grows large.

server/src/controllers/user.ts
```diff
@@ -3,99 +3,108 @@
...
-          fi: '',
-          sv: '',
-          en: ''
-        }
+export async function getUserCourses(req: Request, res: Response): Promise<void> {
+  try {
+    const teacherCourses: TeacherCourses = { current: [], previous: [] };
+
+    // TODO: Go through course_role instead
+    let instances: Array<CourseInstance> = await models.CourseInstance.findAll({
```
jj-stigell, line 22: We could query all courses belonging to the teacher here, and avoid querying inside the loop later on.
For an example with:

```typescript
let instances: Array<Course> = await models.Course.findAll({
  attributes: ['id', 'courseCode', 'minCredits', 'maxCredits'],
  include: [{
    model: CourseInstance,
    where: {
      responsibleTeacher: Number(req.params.userId)
    },
    order: ['startDate', 'DESC']
  },
  {
    model: CourseTranslation,
    attributes: ['language', 'courseName', 'department'],
  }],
});
```

Should fetch all courses, translations and their instances (arranged by start date descending) with teacher id. Then it would be just a matter of reformatting the data to match the model we want to return to the user.

server/src/controllers/user.ts
```diff
@@ -3,99 +3,108 @@
...
+      where: {
+        responsibleTeacher: Number(req.params.userId)
```
jj-stigell, line 24: This is just a note for the future. It is better not to put request params directly to the query, rather first run validation on it.

jj-stigell (changes requested):
Looking good, just few points regarding the database queries that I would like you to address

server/src/controllers/user.ts
no-comma: Yes, that is much better. It's been a while since I've used SQL (and Sequelize never) so I didn't remember/know how to do this in a better way, thanks!

server/src/controllers/course.ts
jj-stigell: If possible? Having arbitrary numbers as language identifiers can be confusing. Matching with db enums should make querying easier when querying for example course translations.

jj-stigell (approved):
Good job!

Tonits (approved):

PR #79, opened 2022-12-10
set sequences to restart from 1 after seed:down

jj-stigell:
Sets sequence generator start value back to 1 after running seed:down


Gskartwii (approved):
Perfect, much better than my TRUNCATE idea

no-comma (approved):

PR #81, opened 2022-12-11
Bump up password storage size

Gskartwii:
When writing some test code for Argon2id, I found that the ASCII representation of the hash could become much larger than the current 60, even up to 100 characters long. Let's make the DB field 255 characters long to be safe.


jj-stigell (approved):
Good point. I had a simple bcrypt implementation on my mind when writing this.

no-comma (approved):

PR #111, opened 2023-01-11
Docker-compose with entrypoint

lakooi:
How to use:

- Define postgre password
- Run docker-compose up


server/Dockerfile.test
```diff
@@ -0,0 +1,23 @@
...
+
+FROM node:16-alpine
+
+WORKDIR /app
+
+COPY . .
+
+RUN chown -R node:node ./*
+
+RUN npm install
```
no-comma, line 13: Shouldn't this be `npm ci`

docker-tests/docker-compose.yaml
```diff
@@ -0,0 +1,37 @@
...
+
+  back-end:
+    build: 
+      context: ../server/
+      dockerfile: ../server/Dockerfile.test
+    depends_on:
+      - database
+    ports:
+    - "3000:3000"
+    container_name: backend-server
```
no-comma, line 17: I run into a naming conflict when I've also used the Docker Compose at the root of the project. This should probably be renamed.

server/Dockerfile.test
```diff
@@ -0,0 +1,23 @@
...
+
+RUN npm install --save-dev sequelize-cli
```
no-comma, line 15: Why are you installing `sequelize-cli` separately?

server/docker-run-tests.sh
```diff
@@ -0,0 +1,13 @@
+# SPDX-FileCopyrightText: 2022 The Aalto Grades Developers
+#
+# SPDX-License-Identifier: MIT
+
+#!/bin/sh
+
+cat build/configs/database.js
```
no-comma, line 7: This path is wrong, it should be `build/src/configs/database.js`. But why is this here?

no-comma (changes requested):

lakooi:
I noticed that the original docker-compose file for backend contained the same scripts as in my entrypoint.sh, so I am thus removing the custom dockerfile/bash file.

no-comma:
Remember to update the year on new files to 2023.

docker-tests/docker-compose.yaml
lakooi: Fixed

server/Dockerfile.test
```diff
@@ -0,0 +1,21 @@
...
+
+RUN chown -R node:node ./*
+
+RUN npm ci
+
+RUN npm run build
+
+USER node
+
+CMD ["/bin/sh", "-c", "npm run migration:up && npm run seed:up && npm test"]
```
no-comma, line 19: Is it possible to pass some argument to specify either test or start so we don't need two nearly identical Dockerfiles?

no-comma (commented):
I think this is good now, but I won't approve it because it also contains changes from me.

Gskartwii (approved):
LGTM! Just curious, what's the purpose of the conditionality of `npm ci`?

no-comma:
> LGTM! Just curious, what's the purpose of the conditionality of `npm ci`?

The main idea is to avoid unnecessarily running `npm ci` twice in the PR workflow in GitHub. Once outside the test container so the linter can be ran and then again while creating the test container.

jj-stigell (approved):
Looking good. Added fix to the test rerun issue by adding "seeds" table to the database that keeps track of seed. Sequelize will no skip loading seeds that have been loaded on previous test runs.

PR #112, opened 2023-01-11
Individual course and course instance route

jj-stigell:
implements #68 

Also added foreign keys to the sequelize model init to fix some querying issues 866f90d25ef714a77377d9b7b4b3efd9b89c7eda


server/src/controllers/course.ts
```diff
@@ -33,4 +57,152 @@ export async function addCourse(req: Request, res: Response): Promise<void> {
...
-}
\ No newline at end of file
+}
+
+export async function getCourse(req: Request, res: Response): Promise<Response> {
+  try {
+    const courseId: number = Number(req.params.courseId);
+    
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const course: any = await courseService.findCourseById(courseId, null);
```
no-comma, line 67: Are you using `any` to access other included models without `@ts-ignore`? If so we should discuss which way we want to handle this.

server/src/services/user.ts
```diff
@@ -0,0 +1,11 @@
+// SPDX-FileCopyrightText: 2022 The Aalto Grades Developers
```
no-comma, line 1: 2023

server/src/services/index.ts
```diff
@@ -0,0 +1,6 @@
+// SPDX-FileCopyrightText: 2022 The Aalto Grades Developers
```
no-comma, line 1: 2023

server/src/services/course.ts
```diff
@@ -0,0 +1,35 @@
+// SPDX-FileCopyrightText: 2022 The Aalto Grades Developers
```
no-comma, line 1: 2023

server/test/routes/index.test.ts
```diff
@@ -0,0 +1,81 @@
+// SPDX-FileCopyrightText: 2022 The Aalto Grades Developers
```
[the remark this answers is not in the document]
jj-stigell: updated 9b5462d73d28578c40a82a9569ef4a9d20354ceb

server/src/controllers/course.ts
```diff
@@ -20,6 +23,27 @@ export interface Course {
   evaluationInformation: LocalizedString
 }
 
+export interface Instance extends Course {
```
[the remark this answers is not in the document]
jj-stigell: Agree. Interface names updated to differentiate between models and interfaces. 827b0aeee12d29ecd01d60467dba17517903ff52

server/test/routes/index.test.ts
```diff
@@ -0,0 +1,81 @@
...
+    expect(res.body.course.minCredits).toBeDefined();
+    expect(res.body.course.maxCredits).toBeDefined();
+    expect(res.body.course.department).toBeDefined();
+    expect(res.body.course.name).toBeDefined();
+    expect(res.body.course.evaluationInformation).toBeDefined();
+    expect(res.statusCode).toBe(200);
+  });
+
+  it('should respond with 404 not found, if non-existing course id', async () => {
+    const res: supertest.Response = await request.get('/v1/courses/9999999999999');
```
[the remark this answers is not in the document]
jj-stigell: Works, good point! Number 0 should also be good. Postgresql integer is signed but we have the PKs set as an incrementing sequence starting from 1, so there should never be a situation were -1 or 0 is PK. As long as we don't add input validation that checks that id is positive integer, the '-1' will work in this test.

Updated it to -1 to make reading the code more pleasant for the eyes f06244e7b0667cebde76de5ab32d7e2a32fc3b11  👀

server/src/services/course.ts
```diff
@@ -0,0 +1,35 @@
...
+//
+// SPDX-License-Identifier: MIT
+
+import { IncludeOptions } from 'sequelize';
+import models from '../database/models';
+import Course from '../database/models/course';
+import CourseInstance from '../database/models/courseInstance';
+import CourseTranslation from '../database/models/courseTranslation';
+
+export async function findCourseById(courseId: number, instanceId: number | null): Promise<Course> {
```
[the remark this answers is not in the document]
jj-stigell: added comments to services ec49c5c448f9ebfd1654a9e2d3e41ab5addf90c9

server/src/controllers/course.ts
jj-stigell: Yes. I tried alternative solution for this problem by extending the course model in the services to have translation and instance array. What do you think about this approach? 7d1f7c83268caf8efca11f005b70ba078329b07b

no-comma: Of the three options mentioned here I think that might be the best one. `any` and `@ts-ignore` seem hacky and to an extent defeat the purpose of using TypeScript.

no-comma (approved):
LGTM

server/src/controllers/course.ts
```diff
@@ -20,6 +23,16 @@ export interface CourseData {
   evaluationInformation: LocalizedString
 }
 
+export interface InstanceData extends CourseData {
```
Gskartwii, line 26: Is the purpose of this interface to match what Sisu returns, or is it rather meant for our internal representation purposes? If it's the latter, I would prefer to use composition over inheritance here.

server/src/controllers/course.ts
```diff
@@ -40,3 +53,146 @@ export async function addCourse(req: Request, res: Response): Promise<void> {
...
+  
+    res.status(200);
+    return res.send({
+      success: true,
+      course: courseData
+    });
+  } catch (error: unknown) {
+    console.log(error);
+  
+    if (error instanceof Error) {
```
Gskartwii, line 109: Could we have more fine-grained checking of the kind of Error here? I think a database error, for instance, shouldn't cause a 404, and that's not guaranteed very strongly here.

server/src/controllers/course.ts
```diff
@@ -40,3 +53,146 @@ export async function addCourse(req: Request, res: Response): Promise<void> {
...
+    return res.send({
+      success: false,
+      error: 'Internal Server Error'
+    });
+  }
+}
+      
+export async function getInstance(req: Request, res: Response): Promise<Response> {
+  try {
+    const courseId: number = Number(req.params.courseId);
```
Gskartwii, line 126: What if this produces a `NaN`?

server/test/routes/index.test.ts
```diff
@@ -1,3 +1,4 @@
+// SPDX-FileCopyrightText: 2023 The Aalto Grades Developers
```
Gskartwii, line 1: Should we combine these two lines by writing `2022 - 2023`?

Gskartwii (changes requested):

server/test/routes/index.test.ts
no-comma: My suggestion for how to handle the copyright year on our files was to just include the year of the original creation of that file, 2022 on files created in 2022, 2023 on files created in 2023, and so on. Using a range is equally as valid, but we should be consistent. Considering that this project will eventually have other people developing it, keeping consistency is probably easier with just having the year of file creation.

server/src/controllers/course.ts
jj-stigell: you mean using classes instead of interface?

server/test/routes/index.test.ts
jj-stigell: line removed

server/src/controllers/course.ts
no-comma: I think he means having the `CourseData` be a member of `InstanceData` rather than `InstanceData` extending it:
```
interface InstanceData {
  courseData: CourseData,
  ...
}
```

jj-stigell: > I think he means having the `CourseData` be a member of `InstanceData` rather than `InstanceData` extending it:
> 
> ```
> interface InstanceData {
>   courseData: CourseData,
>   ...
> }
> ```

Follow up question. What if we would use class here?
Something like:
```JavaScript
// class for course
class CourseData {
  id: number;
  courseCode: string;
  minCredits: number;
  maxCredits: number;
  department: LocalizedString;
  name: LocalizedString;
  evaluationInformation: LocalizedString;

  constructor(course: Course) {
    this.id = course.id;
    this.courseCode = course.courseCode;
    this.minCredits = course.minCredits;
    this.maxCredits = course.maxCredits;
    // + Other data parser for translations, instances etc.
  }
}

// route function
const course: Course = await courseService.findCourseById(courseId);
const parsedCourse: CourseData = new CourseData(course);
````

Different branches implementing different routes are now using variety of ways to parse the translation and course instance data. At some point we should try to refactor the code for reusability, could a class approach solve this? Or does this make any sense?

Comment (jj-stigell) [reply server/src/controllers/course.ts]: updated https://github.com/aalto-grades/base-repository/pull/112/commits/19a2ea7e85569ff5151f0978bfcf930351875a72

File: server/src/controllers/course.ts (hunk 10 above)
Comment (jj-stigell) [reply server/src/controllers/course.ts]: updated errors https://github.com/aalto-grades/base-repository/pull/112/commits/a3fa4c871221cae93c332da6d710e7631ff39df4

File: server/src/controllers/course.ts (hunk 11 above)
Comment (jj-stigell) [reply server/src/controllers/course.ts]: yup added https://github.com/aalto-grades/base-repository/pull/112/commits/58b3eaa39ed0167afab1113fd9bfd0d68ae98679

**Comment by jj-stigell** [comment]
instance route updated, no course id required when fetching instance data

File: server/src/services/course.ts [hunk 13]
```diff
@@ -10,3 +12,37 @@ export interface CourseWithTranslationAndInstance extends Course {
...
+
+/**
+ * Finds a course by its id, and includes related data from the CourseTranslation and CourseInstance models,
+ * depending on whether an instanceId is provided.
+ * @param {number} courseId - The id of the course to be found.
+ * @param {number | null} [instanceId=null] - The id of the course instance, defaults to null. If provided, it will include data from the CourseInstance model for the instance with that specific id.
+ * @returns {Promise<Course>} - A promise that resolves with the found course object.
+ * @throws {Error} - If the course or course instance is not found, it throws an error with a message indicating the missing course or course instance.
+ */
+export async function findCourseById(courseId: number, instanceId: number | null = null): Promise<CourseWithTranslationAndInstance> {
```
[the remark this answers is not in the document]
jj-stigell: true, removed https://github.com/aalto-grades/base-repository/pull/112/commits/9c6a4ac876161aa782c34562ad57bbb6668632e6

server/src/services/instance.ts
```diff
@@ -0,0 +1,41 @@
...
+//
+// SPDX-License-Identifier: MIT
+
+import models from '../database/models';
+import Course from '../database/models/course';
+import CourseInstance from '../database/models/courseInstance';
+import CourseTranslation from '../database/models/courseTranslation';
+
+export interface InstanceWithCourseAndTranslation extends CourseInstance {
+  Course: Course & {
```
no-comma, line 11: `CourseWithTranslation` should work here now.

server/src/controllers/course.ts
Gskartwii: Awesome, this is what I was going for!

server/src/controllers/course.ts
```diff
@@ -40,3 +62,161 @@ export async function addCourse(req: Request, res: Response): Promise<void> {
...
+    });
+  
+    return res.status(200).send({
+      success: true,
+      course: courseData
+    });
+  } catch (error: unknown) {
+    console.log(error);
+
+    if (error instanceof yup.ValidationError) {
```
Gskartwii, line 118: I like this!

server/src/controllers/course.ts
```diff
@@ -40,3 +62,161 @@ export async function addCourse(req: Request, res: Response): Promise<void> {
...
+      return res.status(400).send({
+        success: false,
+        error: error.errors
+      });
+    }
+
+    if (error instanceof Error && error?.message.startsWith('course with an id')) {
```
Gskartwii, line 125: In the long run, it could be useful to have our own Error classes, for example `CourseNotFoundError`. Then, we could use `instanceof` here, too.

server/src/controllers/course.ts
```diff
@@ -40,3 +62,161 @@ export async function addCourse(req: Request, res: Response): Promise<void> {
...
+    console.log(error);
+
+    if (error instanceof yup.ValidationError) {
+      return res.status(400).send({
+        success: false,
+        error: error.errors
+      });
+    }
+
+    if (error instanceof Error && error?.message.startsWith('course instance with an id')) {
```
Gskartwii, line 210: See comment above

server/src/services/user.ts
```diff
@@ -0,0 +1,17 @@
...
+
+/**
+ * Finds a user by its id.
+ * @param {number} userId - The id of the user to be found.
+ * @returns {Promise<User>} - A promise that resolves with the found user object.
+ * @throws {Error} - If the user is not found, it throws an error with a message indicating the missing user with the specific id.
+ */
+export async function findUserById(userId: number): Promise<User> {
+  const user: User | null = await User.findByPk(userId);
+  if (!user) throw new Error (`user with id ${userId} not found`);
```
Gskartwii, line 15: For instance, here: could we have a new class extending `Error` for this error? I think this is partially up to preference, and the present method could just as well be suitable in the TypeScript world

Gskartwii (commented):
Looking great, some open questions to resolve before I'm fully comfortable with approval

server/src/controllers/course.ts
jj-stigell: 100% agree, it will make implementing the error handler much easier once we have proper error classes. We are just starting to see what kind of errors we will encounter (apart from the obvious ones).

server/src/services/user.ts
jj-stigell: Yes, we could! Don't know if we could utilize any of the [sequelize errors](https://sequelize.org/api/v6/identifiers.html#errors) here, or maybe these are more for the sequelize internals and not for the actual query results. Anyways this needs some careful thinking before implementing.

Gskartwii (approved):
With that, I say we can merge this PR. In some future refactoring, we could look into error handling with custom classes or the like.

PR #116, opened 2023-01-16
Course instance creation

### LINKED ISSUE #102: Route for creating new course instances

### LINKED ISSUE #103: Insert course instance data to database
Related to #102

no-comma:
Closes #102
Closes #103


server/src/controllers/course.ts
```diff
@@ -40,3 +42,118 @@ export async function addCourse(req: Request, res: Response): Promise<void> {
...
+
+enum Period {
+  I = 'I',
+  II = 'II',
+  III = 'III',
+  IV = 'IV',
+  V = 'V'
+}
+
+function validatePeriod(period: any): period is Period {
```
jj-stigell, line 66: Is there a way to avoid using any type?

server/src/controllers/course.ts
```diff
@@ -40,3 +42,118 @@ export async function addCourse(req: Request, res: Response): Promise<void> {
...
+      throw new Error(`Course with ID ${courseId} does not exist`);
+    }
+
+    models.CourseInstance.create({
+      courseId: courseId,
+      gradingType: request.gradingType,
+      startingPeriod: request.startingPeriod,
+      endingPeriod: request.endingPeriod,
+      teachingMethod: request.teachingMethod,
+      responsibleTeacher: request.responsibleTeacher, // TODO: Is this fine?
```
jj-stigell, line 142: Works as long as the responsible teacher exists in our users table. request with non existing id will throw error, for an example "'Key (responsible_teacher)=(-1) is not present in table "user".'". 

Maybe we could check if the teacher actually exists with something like: 
```javascript
const teacher: User | null = await models.User.findOne({ where: { responsibleTeacher: request.responsibleTeacher } });
if (teacher === null) {
  throw new Error(`Responsible teacher with ID ${request. responsibleTeacher} does not exist`);
} else {
  // create a new instance
}
```

server/src/controllers/course.ts
```diff
@@ -40,3 +42,118 @@ export async function addCourse(req: Request, res: Response): Promise<void> {
...
+      period === 'V'
+  );
+}
+
+enum TeachingMethod {
+  Lecture = 'LECTURE',
+  Exam = 'EXAM'
+}
+
+function validateTeachingMethod(teachingMethod: any): teachingMethod is TeachingMethod {
```
jj-stigell, line 81: any type

server/src/controllers/course.ts
```diff
@@ -40,3 +42,118 @@ export async function addCourse(req: Request, res: Response): Promise<void> {
...
+  gradingType: GradingType;
+  startingPeriod: Period;
+  endingPeriod: Period;
+  teachingMethod: TeachingMethod;
+  responsibleTeacher: number;
+  startDate: string;
+  endDate: string;
+}
+
+function validateCourseInstanceAddFormat(body: any): body is CourseInstanceAddRequest {
```
jj-stigell, line 98: any type

server/test/routes/index.test.ts
```diff
@@ -75,3 +75,91 @@ describe('Test GET /v1/user/:userId/courses', () => {
     expect(courses.previous[0].courseCode).toBe('TU-A1100');
   });
 });
+
+describe('Test POST /v1/courses/:courseId/instances', () => {
```
jj-stigell, line 79: should we also test the case if responsible teacher does not exist?

server/.eslintrc.json
```diff
@@ -34,6 +34,7 @@
       "error",
       "always"
     ],
+    "@typescript-eslint/no-explicit-any": "off",
```
jj-stigell, line 37: shouldn't we avoid using the any type?

jj-stigell (dismissed):
Looking good, just few questions

server/test/routes/index.test.ts
no-comma: True! I added a test for that.

server/.eslintrc.json
jj-stigell: I modified the validation to return all encountered errors during validation process.

I made quite a few changes so I think someone else should review these.

server/test/routes/index.test.ts
```diff
@@ -75,3 +75,131 @@ describe('Test GET /v1/user/:userId/courses', () => {
...
+          startingPeriod: 'I',
+          endingPeriod: 'II',
+          teachingMethod: 'LECTURE',
+          responsibleTeacher: 1,
+          startDate: '2022-7-10',
+          endDate: '2022-11-10'
+        });
+
+    expect(res.body.success).toBe(false);
+    expect(res.statusCode).toBe(401);
```
jyrge, line 185: Would HTTP 404 be more appropriate here?

server/test/routes/index.test.ts
```diff
@@ -75,3 +75,131 @@ describe('Test GET /v1/user/:userId/courses', () => {
...
+          startingPeriod: 'I',
+          endingPeriod: 'II',
+          teachingMethod: 'LECTURE',
+          responsibleTeacher: -1,
+          startDate: '2022-7-10',
+          endDate: '2022-11-10'
+        });
+
+    expect(res.body.success).toBe(false);
+    expect(res.statusCode).toBe(401);
```
jyrge, line 203: Again with the status code: 422 perhaps?

jyrge (changes requested):

jyrge (approved):

server/src/controllers/course.ts
```diff
@@ -106,12 +109,130 @@ export async function addCourse(req: Request, res: Response): Promise<void> {
...
+
+    if (error instanceof yup.ValidationError) {
+      res.status(400);
+      return res.send({
+        success: false,
+        error: error.errors
+      });
+    }
+
+    res.status(401);
```
Gskartwii, line 222: Maybe we could consider returning a 404 for some of the errors, if they point to a course/teacher not having been found. A 500 could also be suitable for database errors.

server/src/controllers/course.ts
```diff
@@ -106,12 +109,130 @@ export async function addCourse(req: Request, res: Response): Promise<void> {
...
+      throw new Error(`Course with ID ${courseId} does not exist`);
+    }
+
+    const teacher: User | null = await models.User.findOne({
+      where: {
+        id: request.responsibleTeacher
+      },
+    });
+
+    // TODO: Also check that user has correct role
```
Gskartwii, line 192: Could this be best done in a middleware?

server/src/controllers/course.ts
```diff
@@ -106,12 +109,130 @@ export async function addCourse(req: Request, res: Response): Promise<void> {
...
+      endingPeriod: request.endingPeriod,
+      teachingMethod: request.teachingMethod,
+      responsibleTeacher: request.responsibleTeacher,
+      startDate: request.startDate,
+      endDate: request.endDate
+    });
+
+    return res.send({
+      success: true,
+      instance: newInstance
```
Gskartwii, line 210: I wonder what kind of data is sent to the frontend here. It gives the impression that the data is just echoed here, which feels slightly redundant.

Gskartwii (changes requested):
Great job on the clean code and formatting. I had a few questions about the status codes and other details.

server/src/controllers/course.ts
jj-stigell: I made this edit, so I'll comment. Yes basically its bit redundant since it returns all the data back to client. Only new data included is the instance `id` in db and `createdAt` and `updatedAt` values. If client does not need these values then we could just return the `success: true` or just include the data client does not have:

```JavaScript
return res.send({
  success: true,
  instance: {
    id:  newInstance.id
  }
})
```

server/src/controllers/course.ts
no-comma: You're right, the status codes used here are not ideal. @jyrge also commented on this, and we thought that it would be best to discuss our API specs and error handling at the beginning of the next sprint and then go back to update our old routes. So far we've all been doing things in slightly different ways, so getting some collective consensus going forward is desirable.

server/src/controllers/course.ts
no-comma: Actually, this check shouldn't be done in the first place since course roles are instance specific. My thought process for adding this comment was that we wouldn't accidentally want to add a student as a teacher for a course for example, but I'll remove this comment now.

server/src/controllers/course.ts
no-comma: I just noticed that this is returning a Sequelize model too, do you think any issues could arise from doing that?

server/src/controllers/course.ts
Gskartwii: +1 for the consensus. With that, this code should be good for now.

server/src/controllers/course.ts
Gskartwii: Looking at the [Sequelize docs](https://sequelize.org/api/v6/class/src/model.js~model), it would seem that the clas has some properties such as `isNewRecord` and `sequelize`, and overall I just get the idea that it's not intended for serialization as-is. I feel it would be best to only return the id to the client, perhaps also the creation/update date. What are the requirements of the frontend side?

Gskartwii (changes requested):
Let's return something other than the Sequelize model instance to the client.

server/src/controllers/course.ts
no-comma: I changed it to return just the ID of the new instance as suggested above, in case the client needs more information this can be updated later.

Gskartwii (approved):
Awesome!

Issue #121, opened 2023-01-17
Fetching parsed SISU data


no-comma:
As a frontend developer, I want to have an API to fetch parsed SISU contents of a course, so that I can visualize it in an edit page.

- [x] #108 
- [x] #109 
- [x] #110

Issue #120, opened 2023-01-17
Fetching course information from SISU


no-comma:
As a backend developer, I want to get course information from SISU with an API key, so I can parse the content later for the frontend.

- [x] #106 
- [x] #107 
- [x] Testing in CI/CD

lakooi:
- Get better SISU api key for organization instead of personal

no-comma:
For testing we're just going to mock the SISU API.

PR #126, opened 2023-01-22
Fetch and parse course instance data from Sisu

jj-stigell:
implements:
#121
#120
#110
#107 

NOTE. for now , in case of error (bad request, non-existing instance/course id, etc.), will return only 500 "Internal Server Error". 

Teaching periods in Sisu seem to be saved in format "2020-2021 Spring III-V, 2021-2022 Spring III-V". If we want to return periods in format:
```JSON
{
  "startingPeriod": "III",
  "endingPeriod": "V"
}
```
maybe we have to do some kind of  function (using regex?) that filters out the roman numerals from the string? Now starting and ending periods are just returned as '-'.


server/src/controllers/course.ts
```diff
@@ -40,3 +86,60 @@ export async function addCourse(req: Request, res: Response): Promise<void> {
...
+    const parsedInstances: Array<InstanceData> = instancesFromSisu.data.map((instance: SisuInstance) => parseSisuInstance(instance));
+
+    return res.status(200).send({
+      success: true,
+      instances: parsedInstances
+    });
+  } catch (error: unknown) {
+    console.log(error);
+
+    return res.status(500).send({
```
jyrge, line 111: Should the status code / response body be more specific about what went wrong? E. g. 404 if the course was not found.

server/src/controllers/course.ts
```diff
@@ -40,3 +86,60 @@ export async function addCourse(req: Request, res: Response): Promise<void> {
...
+
+    return res.status(200).send({
+      success: true,
+      instance: instance
+    });
+
+  } catch (error: unknown) {
+    console.log(error);
+
+    return res.status(500).send({
```
jyrge, line 140: Same here

server/test/routes/indext.test.ts
```diff
@@ -0,0 +1,91 @@
+// SPDX-FileCopyrightText: 2022 The Aalto Grades Developers
+//
+// SPDX-License-Identifier: MIT
+
```
jyrge, line 4: Typo in file name? "indext.test.js", should it be "index.test.js"?

jyrge (changes requested):

jyrge (approved):

server/src/controllers/course.ts
```diff
@@ -20,12 +24,54 @@ export interface CourseData {
...
+
 export enum Language {
   English = 'EN',
   Finnish = 'FI',
   Swedish = 'SV'
 }
 
+function parseSisuInstance (instance: SisuInstance): InstanceData {
+  return {
+    // instance id here is Sisu instance id (string)
```
no-comma, line 45: This should also be specified as a comment in `CourseData` that the ID can either represent our internal ID (number) or Sisu ID (string).

server/src/controllers/course.ts
```diff
@@ -20,12 +24,54 @@ export interface CourseData {
...
+      en: instance.summary.assesmentMethods.en,
+      fi: instance.summary.assesmentMethods.fi,
+      sv: instance.summary.assesmentMethods.sv
+    },
+    startingPeriod: '-',
+    endingPeriod: '-',
+    startDate: instance.startDate,
+    endDate: instance.endDate,
+    courseType: instance.type === 'exam-exam' ? 'EXAM' : 'LECTURE',
+    gradingType: instance.summary.gradingScale.fi === '0-5' ? 'NUMERICAL' : 'PASSFAIL',
```
no-comma, line 70: This may be fine for now but we should (maybe in the next sprint) use the enums from #116. I would add a TODO reminder to fix this.

server/src/controllers/course.ts
```diff
@@ -20,12 +24,54 @@ export interface CourseData {
...
+  responsibleTeachers: Array<string>,
+}
+
 export enum Language {
   English = 'EN',
   Finnish = 'FI',
   Swedish = 'SV'
 }
 
+function parseSisuInstance (instance: SisuInstance): InstanceData {
```
no-comma, line 43: Extra space.

no-comma (changes requested):

server/test/routes/index.test.ts
```diff
@@ -75,3 +79,82 @@ describe('Test GET /v1/user/:userId/courses', () => {
...
+    expect(res.body.instances[0].endingPeriod).toBeDefined();
+    expect(res.body.instances[0].startDate).toBeDefined();
+    expect(res.body.instances[0].endDate).toBeDefined();
+    expect(res.body.instances[0].courseType).toBeDefined();
+    expect(res.body.instances[0].gradingType).toBeDefined();
+    expect(res.body.instances[0].responsibleTeachers).toBeDefined();
+    expect(res.status).toEqual(200);
+  });
+
+  it('should respond with error when course does not exist/does not have active instances', async () => {
```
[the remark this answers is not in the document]
jj-stigell: I have noticed on few different cases when playing with the Sisu API, that it returns the same error code and message for both non-existing courses and courses with no active instances:
```JSON
{
  "error": {
    "code": 102,
    "message": "Course unit realisations not found"
  }
}
```

I created two separate tests, one for invalid course code and one for no active instances https://github.com/aalto-grades/base-repository/pull/126/commits/9ac4679dd60834363fd7724fdffc07e634e050af. Until we get more detailed documentation concerning the Sisu API, I think this is the best I can do from testing point of view.

server/src/controllers/course.ts
jj-stigell: added reminder https://github.com/aalto-grades/base-repository/pull/126/commits/d9bb67b83c330f34b6a651500784ddca26586d13

server/src/controllers/course.ts
jj-stigell: fixed

server/src/controllers/course.ts
jj-stigell: I noticed that the Sisu API does not have/return unique id for the course itself, just for the course instance. Therefore I edited the `CourseData` interface `id` so it can be undefined when representing parsed sisu course data. What do you think? https://github.com/aalto-grades/base-repository/pull/126/commits/af44fb7adac4d0c89819dc85039d07bb5c87b9d7

no-comma: That's fine too as long as it's explained why the ID can be undefined, which is explained.

no-comma (approved):

PR #127, opened 2023-01-22
Implement route for creating a new course

### LINKED ISSUE #63: Route and database connection for creating a new course

jyrge:
Resolves #63


server/src/controllers/course.ts
```diff
@@ -26,17 +28,57 @@ export enum Language {
...
 
+interface JsonError extends Error {
+  expose?: boolean,
+  status?: number,
+  statusCode?: number,
+  body?: string,
+  type?: string
+}
+
+// Middleware function for handling errors in JSON parsing
```
jj-stigell, line 39: This seems to be the first PR implementing a custom middleware, should we put the middlewares to their own folder right from the beginning?

jyrge: Might be a good idea in future. Should we discuss this in the beginning of next sprint, while designing the common error handling more in detail?

server/src/database/migrations/00003-add-constraints.ts
```diff
@@ -9,16 +9,6 @@ export default {
   up: async (queryInterface: QueryInterface): Promise<void> => {
     const transaction: Transaction = await queryInterface.sequelize.transaction();
     try {
-      await queryInterface.addConstraint('course', {
```
no-comma, line 12: This constraint should be added for course instances now.

server/src/controllers/user.ts
```diff
@@ -26,7 +26,7 @@ export async function getUserCourses(req: Request, res: Response): Promise<void>
 
     // TODO: Go through course_role instead
     const courses: Array<CourseWithTranslationAndInstance> = await models.Course.findAll({
-      attributes: ['id', 'courseCode', 'minCredits', 'maxCredits'],
+      attributes: ['id', 'courseCode' ],
```
no-comma, line 29: Extra space in brackets.

server/src/services/instance.ts
```diff
@@ -21,10 +21,10 @@ export interface InstanceWithCourseAndTranslation extends CourseInstance {
 export async function findInstanceById(instanceId: number): Promise<InstanceWithCourseAndTranslation> {
 
   const instance: CourseInstance | null = await models.CourseInstance.findByPk(instanceId, {
-    attributes: ['id', 'gradingType', 'startingPeriod', 'endingPeriod', 'teachingMethod', 'startDate', 'endDate', 'responsibleTeacher'],
+    attributes: ['id', 'gradingType', 'startingPeriod', 'endingPeriod', 'teachingMethod', 'minCredits', 'maxCredits', 'startDate', 'endDate', 'responsibleTeacher'],
     include: {
       model: Course,
-      attributes: ['id', 'courseCode', 'minCredits', 'maxCredits'],
+      attributes: ['id', 'courseCode' ],
```
no-comma, line 27: Extra space.

server/src/controllers/course.ts
no-comma: I think we should discuss this in the next sprint and decide on uniform conventions to follow.

server/src/controllers/course.ts
```diff
@@ -90,19 +101,105 @@ function parseSisuInstance(instance: SisuInstance): InstanceData {
...
+      errors: [ `SyntaxError: ${err.message}: ${err.body}` ]
+    });
+  } else
+    next();
+}
 
 export async function addCourse(req: Request, res: Response): Promise<void> {
+  const requestSchema: yup.AnyObjectSchema = yup.object().shape({
+    courseCode: yup.string().required(),
+    translations: yup.array().of(yup.object().shape({
```
no-comma, line 118: Since we're already returning the department and name in the `LocalizedString` style, we should make it so POST requests also require the same style:
```
{
  courseCode: 'ELEC-A7200',
  name: {
    en: 'Signals and Systems',
    fi: 'Signaalit ja järjestelmät',
    sv: ''
  },
  department: {
    en: 'School of Electrical Engineering',
    fi: 'Sähkötekniikan korkeakoulu',
    sv: ''
  }
};
```

no-comma (changes requested):

server/src/controllers/course.ts
```diff
@@ -90,19 +101,105 @@ function parseSisuInstance(instance: SisuInstance): InstanceData {
...
+      lang: yup.mixed<Language>().oneOf(Object.values(Language)).required(),
+      department: yup.string().required(),
+      name: yup.string().required()
+    })).required()
+  });
+
+  const t: Transaction = await sequelize.transaction();
```
jj-stigell, line 125: Remember to pass the [transaction](https://sequelize.org/docs/v6/other-topics/transactions/#unmanaged-transactions) to those db calls you want to use it. Now none of the calls utilize this and commit/rollback will do nothing.

server/src/database/models/course.ts
```diff
@@ -8,8 +8,6 @@ import { sequelize } from '..';
 export default class Course extends Model<InferAttributes<Course>, InferCreationAttributes<Course>> {
   declare id: CreationOptional<number>;
   declare courseCode: string;
```
jj-stigell, line 10: Not concerning the changes you made but should we enforce `unique` in the `courseCode`? While testing I could create multiple courses with same `courseCode`. This could create issues once we start adding course instances. To which course should we assign the instances if there are multiple course entries with the same `courseCode`? 

If we query instances based on the courses `id` it would only return instances signed to that `id`. If we query with `courseCode`, then it would return all instances no matter what the `id` is.

server/src/database/migrations/00003-add-constraints.ts
jj-stigell: Agree. Even tough we have yup validating this in the backend, I would still have the database level check also.

jj-stigell (changes requested):
Looking good, few points to address.

server/src/database/migrations/00003-add-constraints.ts
jj-stigell: Agree. Even tough we have yup validating this in the backend, I would still have the database level check also.

server/src/controllers/course.ts
jj-stigell: fair enough

server/src/controllers/course.ts
```diff
@@ -90,19 +101,105 @@ function parseSisuInstance(instance: SisuInstance): InstanceData {
...
+        sv: ''
+      },
+      evaluationInformation: {
+        en: '',
+        fi: '',
+        sv: ''
+      }
+    };
+
+    for (const translation of req.body.translations) {
```
jj-stigell, line 154: Instead of using db calls in the for loop to create the translations, could we utilize the sequelize [bulkCreate](https://sequelize.org/api/v6/class/src/model.js~model#static-method-bulkCreate)?

Could be maybe something like:
```JavaScript
await CourseTranslation.bulkCreate([
  { /*FI translations*/ },
  { /*EN translations*/ },
  { /*SV translations*/ }
])
```

I think the performance of this route is not so critical because new courses are not added that frequently. In this case looping few database calls would not pose any performance issues (+ its run in a transaction anyway). Its just good to note that sequelize has this kind of method also.

I'll let you decide what to do.

server/src/database/models/course.ts
no-comma: This goes back to the question of whether course codes are unique at Aalto to begin with. I recall not getting a definitive answer. The immediate reaction was to assume that they are unique, but there might have been some cases where they are not with courses that have multiple language versions of them.

We could triple check this again, but I think sticking to using our internal numeric course ID for all logic is the safest option in any case if something changes in the future.

jj-stigell (commented):
One comment concerning sequelize.

server/src/controllers/course.ts
jyrge: > Since we're already returning the department and name in the `LocalizedString` style, we should make it so POST requests also require the same style:
> 
> ```
> {
>   courseCode: 'ELEC-A7200',
>   name: {
>     en: 'Signals and Systems',
>     fi: 'Signaalit ja järjestelmät',
>     sv: ''
>   },
>   department: {
>     en: 'School of Electrical Engineering',
>     fi: 'Sähkötekniikan korkeakoulu',
>     sv: ''
>   }
> };
> ```

I was first considering this style in request. Then I thought that how to validate that the translation for one language is not missing a field. I found out that it would be a bit tricky to implement it with yup. Do you have suggestions?

If that kind of validation is not so critical, I can change the request style right away.

server/src/database/models/course.ts
jyrge: If I remember correct, they said that we can not assume that they are unique, because there may be exceptions to the rule. That's why we finally ended up using a separate, auto-incrementing numeric id for courses.

server/src/controllers/course.ts
jyrge: Oh fuck, true... Will fix this.

jyrge: We probably have to figure out a more thorough testing solution for db operations in future. Not directly concerning this code snippet though.

server/src/controllers/course.ts
jyrge: I could probably do the db operations in bulkCreate and then loop through the array to do the rest of the operations.

server/src/controllers/course.ts
no-comma: Yup allows for defining custom validation methods.
https://github.com/jquense/yup#addmethodschematype-schema-name-string-method--schema-void

We could define a method for `yup.object` which checks for the fields of `LocalizedString`, this could then be used elsewhere too.

server/src/database/migrations/00003-add-constraints.ts
jyrge: True, have probably missed this.

server/src/database/models/course.ts
jj-stigell: > The immediate reaction was to assume that they are unique, but there might have been some cases where they are not with courses that have multiple language versions of them.

Shouldn't the different language versions be defined in instance level, not in course level?

But anyway, I think we should triple check this thing with PO.

no-comma: But as mentioned initially too, this discussion doesn't particularly concern the PR, so it shouldn't block it from being merged and I think can be marked as resolved.

server/src/controllers/course.ts
jyrge: I thought about this kind of scenario: _name_ and _department_ are translated into two languages, utilizing localized strings. If English translation of _department_ is missing but other English translations exist, fail the validation.

Maybe I'm just overthinking this.

no-comma: That might be reasonable but I'm not sure if it's necessary, there could also be a scenario such as:
```
{
  courseCode: 'CS-A1120',
  name: {
    en: 'Programming 2',
    fi: '',
    sv: ''
  },
  department: {
    en: 'Department of Computer Science',
    fi: 'Tietotekniikan laitos',
    sv: 'Institutionen för datateknik'
  }
};
```
I think that this is fine and should pass validation, though the Finnish and Swedish names are missing while the department exists.

jyrge: True

jyrge: Maybe a trivial question, but should we then require the field with empty string or allow it to be left undefined?

no-comma: I would say that it should still be defined to match the `LocalizedString` interface definition, but not sure if it matters.

jyrge: Implementing this so that undefined translations are replaced with empty string when inserting them into database. Yup schema for localized string checks that the object has at least one translation and no translations for any other languages except the ones supported.

server/test/routes/index.test.ts
```diff
@@ -161,6 +157,117 @@ describe('Test GET /v1/user/:userId/courses', () => {
...
+      name: {}
+    };
+    const res: supertest.Response = await request.post('/v1/courses').send(input);
+    expect(res.statusCode).toBe(400);
+    expect(res.body.success).toBe(false);
+    expect(res.body.errors).toContain('department must contain at least one translation');
+    expect(res.body.errors).toContain('name must contain at least one translation');
+  });
+
+  it('should respond with validation errors, if unknown translations are present in a localized string', async () => {
```
no-comma, line 241: These localized string tests are more general than just this route, and they should be tested separately.

server/src/controllers/course.ts
```diff
@@ -157,6 +281,14 @@ const courseInstanceAddRequestSchema: yup.AnyObjectSchema = yup.object().shape({
   responsibleTeacher: yup
     .number()
     .required(),
+  minCredits: yup
+    .number()
+    .min(0)
```
no-comma, line 286: Should this be 1 instead? Is there a course that can give zero credits?

server/src/controllers/course.ts
```diff
@@ -93,19 +118,116 @@ function parseSisuInstance(instance: SisuInstance): InstanceData {
...
-    // TODO: add the course to the database
-    res.send({
-      success: true
+    await requestSchema.validate(req.body, { abortEarly: false });
+
+    const course: Course = await Course.create({
+      courseCode: req.body.courseCode
+    }, { transaction: t });
+
+    const courseTranslations: CourseTranslation[] = await CourseTranslation.bulkCreate([
```
no-comma, line 148: Unless I'm forgetting something, we've used `Array<T>` everywhere else for array type annotations, we should be consistent. This could also be enforced by the linter.

no-comma (changes requested):
Great that you already updated the course instance creation to include min and max credits! Just a couple more things to address.

server/src/controllers/course.ts
jj-stigell: +1 for enforcing one style with linter

server/src/controllers/course.ts
jj-stigell: I was also thinking about this and did some Sisu research. There exists courses with 0 credits like the bachelors 'Maturity Test (0 cr)'. Though I am not sure if that kind of course would ever end up in this system? Nevertheless I think it's better to have the option for 0 credits.

jj-stigell (approved):
LGTM

server/src/controllers/course.ts
no-comma: If the possibility for a 0 credit course exists then I agree that it's better to keep this as 0.

server/test/routes/index.test.ts
jyrge: I could probably create test/controllers/course.test.ts module where to create unit tests for functions in that module.

jyrge: We should also discuss more about our testing strategy in next sprint. And also refactoring the code so that commonly reusable operations would be separated into their own functions (maybe even in separate modules?) and tested then separately as well.

server/src/controllers/course.ts
jyrge: Now enforced by eslint

no-comma (approved):

PR #128, opened 2023-01-25
Routes course instances

### LINKED ISSUE #67: Route and database connection for getting all instances of a course

Tonits:
closes #67


server/src/controllers/course.ts
```diff
@@ -150,3 +159,36 @@ export async function fetchInstanceFromSisu(req: Request, res: Response): Promis
...
+    
+export async function getAllCourseInstances(req: Request, res: Response): Promise<Response> {
+  try {
+    const courseId: number = Number(req.params.courseId);
+    await idSchema.validate({ id: courseId });
+    const instances: Array<CourseInstance> = await courseService.findAllInstances(courseId);
+
+    return res.status(200).send({
+      success: true,
+      instances: instances,
```
no-comma, line 171: We don't want to return Sequelize models, you should construct `InstanceData` objects (already in main) out of all the models and return an array of those.

server/src/services/course.ts
```diff
@@ -10,3 +11,23 @@ export interface CourseWithTranslationAndInstance extends Course {
   CourseTranslations: Array<CourseTranslation>
   CourseInstances: Array<CourseInstance>
 }
+
+export async function findAllInstances(courseId: number): Promise<Array<CourseInstance>> {
+  const course: Array<Course> | null = await models.Course.findAll({
```
no-comma, line 16: `findOne` would be more appropriate, AFAIK `findAll` will continue to search through courses even when we have already found the one we want, whereas `findOne` will return the first (and only) found one right away.

no-comma (changes requested):
Looks good for the most part, just a couple of things to note.

server/src/services/course.ts
jj-stigell: If we are querying here with the course PK, I would recommend using `findByPk`, then there is no need to use the `where` clause. It would also return just one Course instance. More detailed info [here](https://sequelize.org/docs/v6/core-concepts/model-querying-finders/#findbypk).

It could look something like this:
```Javascript
const course: Course | null = await models.Course.findByPk(courseId);

if (!course) throw new Error (`course with id ${courseId} not found`);
```

no-comma: True, that slipped my mind.

server/src/services/course.ts
```diff
@@ -31,22 +33,35 @@ export async function findCourseById(courseId: number): Promise<CourseWithTransl
...
       courseId: courseId
     }
   });
 
-  return instances;
+  const instancesWithTeacher: Array<InstanceWithTeacher> = [];
+
+  if (instances.length != 0) {
+    for (const instance of instances) {
+      const teacher: User | null = await User.findByPk(instance.responsibleTeacher);
```
jj-stigell, line 58: can you use `include` in the `CourseInstance` db query so there is no need for this loop?

server/src/services/course.ts
```diff
@@ -31,22 +33,35 @@ export async function findCourseById(courseId: number): Promise<CourseWithTransl
...
-    where: {
-      id: courseId
-    }
-  });
+export interface InstanceWithTeacher {
+  CourseInstance: CourseInstance
+  Teacher: User | null
+}
+
+export async function findAllInstances(courseId: number): Promise<Array<Course | Array<InstanceWithTeacher>>> {
```
jj-stigell, line 41: Would it be better to return a tuple here? For example:
```JavaScript
Promise<[Course, Array<InstanceWithTeacher>]>
```

Current one seems to imply the course data is also an array or we are repeating course data.

This tuple would be easy to destructure in the route function:
```JavaScript
const [ course, instances ]: [Course, Array<InstanceWithTeacher>] = await courseService.findAllInstances(courseId);
```

jj-stigell (changes requested):
Few things to point out

no-comma:
I'll try to review this again ASAP.

no-comma (commented):
I echo the comments of @jj-stigell, otherwise just a small naming detail I noticed.

server/src/services/course.ts
jj-stigell: Sequelize [documentation](https://sequelize.org/docs/v6/core-concepts/assocs/#defining-an-alias) on using alias when there exists more than one association between two models

Tonits:
This route must be changes in the future to search for `ResponsibleTeacher` through `CourseRole` instead of the foreign key in `CourseInstance`

jyrge (changes requested):

server/test/controllers/courseInstance.test.ts
```diff
@@ -231,3 +231,37 @@ describe('Test GET /v1/instances/:instanceId', () => {
...
+
+  it('should respond with error if course does not exist', async () => {
+    const res: supertest.Response = await request.get('/v1/courses/-1/instances');
+    expect(res.statusCode).toBe(404);
+    expect(res.body.success).toBe(false);
+  });
+
+  it('should respond with error if courseId is not a number', async () => {
+    const res: supertest.Response = await request.get('/v1/courses/a/instances');
+    expect(res.statusCode).toBe(400);
```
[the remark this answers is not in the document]
jyrge: Marking as resolved, as this is not directly related to this PR.

jyrge (approved):

PR #131, opened 2023-01-29
Frontend conditional rendering

kiiamakir:
Added role-based authentication to the private route. This will take care of conditional rendering in the cases where some user roles shouldn't have access to a page. If multiple user groups are able to access a page but its content is should be modified based on the role of the user (for example in the front page) this should be done inside the component.


client/src/components/FrontPage.js
```diff
@@ -10,33 +10,38 @@ import BasicGrid from './front-page/BasicGrid';
...
     coursesService.getCourses()
-      .then((data) => setPreviousCourses(data.courses.previous))
+      .then((data) => {
+        setCurrentCourses(data.courses.current);
+        setPreviousCourses(data.courses.previous);
+      })
       .catch((e) => console.log(e.message));
   }, []);
 
+  // Conditional rendering: only admin and teachers are shown the button for creating a new course
```
viivi-uhari, line 31: The commenting on this PR is very good! So, could this comment be of the same format as the others between components? As an example: CourseView.js, line 35

viivi-uhari (commented):
Great job done with this PR! Didn't see anything major that would need changing.

marillam (approved):
Looks good to me!

kiiamakir:
Conditional rendering still needs to be implemented for the new changes in main

viivi-uhari (approved):
Everything seems good and conditional rendering works!

PR #140, opened 2023-02-02
Implement session expiry in backend

### LINKED ISSUE #105: Session expiry after a given amount of time

### LINKED ISSUE #122: JWT token sessions
As a backend developer, I want to be able to verify the JWT tokens in API requests through middleware, so that the requests are passed correctly if the JWT is correct.

- [ ] #104 
- [ ] #105

Gskartwii:
Closes #122.
Closes #105.


no-comma (commented):
Looks good, though I wonder, is there a particular reason to have 25 hours as the expiry time? Should it maybe be configurable somehow instead?

Gskartwii:
Good question about the hardcoded expiry time. It could be more clear to separate it into a clearly marked constant - or would you even feel it could deserve a new environment variable?

server/src/routes/login.ts
```diff
@@ -65,7 +65,9 @@ export async function authLogin(req: Request, res: Response, next: NextFunction)
               role: loginResult.role,
               id: loginResult.id,
             };
-            const token: string = jwt.sign(body, jwtSecret);
+            const token: string = jwt.sign(body, jwtSecret, {
+              expiresIn: '25h',
```
jj-stigell, line 69: Would it be better to set the `expiresIn` at `configs/config.ts`, now any changes must be applied to multiple places (here and tests)

jj-stigell (changes requested):

no-comma:
> Good question about the hardcoded expiry time. It could be more clear to separate it into a clearly marked constant - or would you even feel it could deserve a new environment variable?

Defining it as a constant in the code would be good enough now I think. I see nothing immediately wrong with 25 hours, just wondering if it has some special significance (e.g. standard session expiry time in multiple services).

jj-stigell (approved):

no-comma (approved):

PR #141, opened 2023-02-06
Swagger setup

Gskartwii:


lakooi:
Note that this PR also removes the default route found in the backend that descibes Hello /something

lakooi:
As I am not very experienced with Typescript I set the the type of swagger JSON objects as : object. Is this the correct typing for these?

Gskartwii:
@lakooi You might be able to find more refined types over here:
- https://www.npmjs.com/package/@types/swagger-ui-express
- https://www.npmjs.com/package/@types/swagger-jsdoc

server/src/routes/swagger.ts
```diff
@@ -0,0 +1,32 @@
+// SPDX-FileCopyrightText: 2022 The Aalto Grades Developers
+//
+// SPDX-License-Identifier: MIT
+
+/**
+ * This file is used as a base for the swagger complition in the index.js file of routes.
+ */
+
+export const definition: object = {
```
Gskartwii, line 9: Is there a more specific type available for this in an `@types` package?

server/src/routes/swagger.ts
```diff
@@ -0,0 +1,32 @@
+// SPDX-FileCopyrightText: 2022 The Aalto Grades Developers
```
Gskartwii, line 1: Year 2023, thanks @no-comma

server/src/routes/index.ts
```diff
@@ -10,30 +10,382 @@ import passport from 'passport';
 import cors from 'cors';
 import cookieParser from 'cookie-parser';
 import { frontendOrigin } from '../configs';
+import { definition } from './swagger';
+
+import swaggerUI from 'swagger-ui-express';
+import swaggerJsdoc from 'swagger-jsdoc';
+
+const options: object = {
```
Gskartwii, line 18: Perhaps a more specific type here, too

Gskartwii (commented):
Nice to have this base here! A couple comments about typing

(Turns out I can't request changes or approve, since I opened this PR)

server/src/routes/swagger.ts
```diff
@@ -0,0 +1,32 @@
...
+//
+// SPDX-License-Identifier: MIT
+
+/**
+ * This file is used as a base for the swagger complition in the index.js file of routes.
```
no-comma, line 6: I would write this as:
`This file is used as a base for the Swagger compilation in routes/index.ts.`

server/src/routes/swagger.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export const definition: object = {
+  openapi: '3.0.1',
+  info: {
+    version: '1.0.0',
+    title: 'APIs Document',
+    description: 'Documentation of Aalto Grades Backend API',
+    termsOfService: '',
+    license: {
+      name: 'MIT',
```
no-comma, line 17: I would write `Expat (MIT)` since this doesn't appear to be an SPDX identifier. https://www.gnu.org/licenses/license-list.html#Expat
> Some people call this license “the MIT License,” but that term is misleading, since MIT has used many licenses for software. It is also ambiguous, since the same people also call the [X11 license](https://www.gnu.org/licenses/license-list.html#X11License) “the MIT License,” failing to distinguish them. We recommend not using the term “MIT License.”

server/src/routes/swagger.ts
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * This file is used as a base for the swagger complition in the index.js file of routes.
+ */
+
+export const definition: object = {
+  openapi: '3.0.1',
+  info: {
+    version: '1.0.0',
+    title: 'APIs Document',
```
no-comma, line 13: Maybe just `API Document`.

server/src/routes/swagger.ts
no-comma, line 1: This file should probably be in `configs` instead.

server/src/routes/index.ts
```diff
@@ -10,30 +10,382 @@ import passport from 'passport';
...
+  apis: ['./src/routes/*.ts'],
+};
+
+const openapiSpecification: object = swaggerJsdoc(options);
 
 export const router: Router = Router();
 
 router.use(cookieParser());
 
+router.use('/api-docs', swaggerUI.serve);
```
no-comma, line 29: I don't see why this line is here, is it necessary?

no-comma (changes requested):

server/src/routes/swagger.ts
no-comma: `swaggerJSDoc.OAS3Definition` seems to be the right type.
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/swagger-jsdoc/index.d.ts

server/src/routes/index.ts
no-comma: `swaggerJSDoc.OAS3Options` I think.
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/swagger-jsdoc/index.d.ts

server/src/routes/index.ts
```diff
@@ -10,30 +10,382 @@ import passport from 'passport';
...
+
+import swaggerUI from 'swagger-ui-express';
+import swaggerJsdoc from 'swagger-jsdoc';
+
+const options: object = {
+  definition,
+  apis: ['./src/routes/*.ts'],
+};
+
+const openapiSpecification: object = swaggerJsdoc(options);
```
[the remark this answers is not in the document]
no-comma: Never mind, this might be intended to be `object`.

server/src/routes/index.ts
lakooi: https://www.npmjs.com/package/swagger-ui-express, according to this documentation

server/src/routes/swagger.ts
lakooi: Changed typing

server/src/routes/index.ts
lakooi: This as well, thanks.

server/src/routes/index.ts
no-comma: `router.use` mounts a middleware function to all requests to a specific path (https://expressjs.com/en/4x/api.html#app.use, this is for `app.use` but `router.use` should work the same way), in this case `swaggerUI.serve` to all requests to `/api-docs`.

In the following line you're also defining `swaggerUI.serve` to be used for GET requests to `/api-docs`, which is redundant. So unless `swaggerUI.serve` is needed for other requests which haven't been explicitly defined here (yet?), then I don't think this line is needed. Otherwise the following line should not include `swaggerUI.serve`.

lakooi: Ok now I understand your point. The router.get function incorrectly contained the swaggerUI.serve function.

Swagger doesn't work without stating router.use for the middleware.

no-comma (approved):
LGTM

jyrge (approved):

PR #143, opened 2023-02-08
Backend file structure refactoring

no-comma:
Documentation of code style and structure will be written separately to the wiki.


server/src/app.ts
```diff
@@ -2,14 +2,16 @@
...
+import cors from 'cors';
 import express, { Application } from 'express';
+
 import { router } from './routes/index';
-import cors from 'cors';
 
 export const app: Application = express();
 
 app.use(cors({
   origin: 'http://localhost:3005',
```
jj-stigell, line 13: instead of hardcoding, should we use the `FRONTEND_ORIGIN` env here?

jj-stigell (changes requested):

server/src/app.ts
no-comma: Yes, good point.

jj-stigell (approved):
👍

no-comma:
This could also split `src/routes/index.ts` into multiple files now that the Swagger PR got merged, but that may be better left to API refactoring PRs.

Tonits (approved):
Looks great!

Issue #136, opened 2023-02-01
Refactoring


no-comma:
As a developer, I want to refactor the code developed in early increments, so that technical debt does not pile up.

- [x] Refactoring client @viivi-uhari 
- [x] Refactoring server file structure @no-comma 
- [x] Documenting server structure @no-comma 
- [x] Middleware for server error handling @jj-stigell  
- [x] Refactoring server routes according to API specification 
  - [x] Auth @Gskartwii 
  - [x] Course @no-comma 
  - [x] Course instances @no-comma 
  - [x] Sisu @no-comma 
  - [x] User @no-comma

Issue #135, opened 2023-02-01
Custom error types in backend


no-comma:
As a developer, I want the system to present good error codes, so errors are clear and don't leak secrets.

PR #146, opened 2023-02-13
Custom error classes

jj-stigell:
#135

This might overlap little bit with the #136 since I implemented simple middleware to test the custom errors.

Http codes are presented as enums to make the code more readable and to present a set of accepted codes. We could add more http codes to the list as we need.

The `CustomError` type extends JS `Error` with new property `HttpCode`. `SisuError` and other custom errors could extend `CustomError` with their own custom properties. This way we can catch the correct error type in the middleware.

`getCourse` function has example implementation of using the new error type:

```JavaScript
if (!course) throw new CustomError(`course with an id ${courseId} not found`, HttpCode.NotFound);
```

Example of how we could throw Sisu API related error:
```JavaScript
throw new SisuError(instanceFromSisu.data.error.code);
```


server/src/controllers/course.ts
```diff
@@ -362,11 +364,24 @@ export async function addCourseInstance(req: Request, res: Response): Promise<Re
...
+
+    const course: CourseWithTranslation | null = await models.Course.findByPk(courseId, {
+      attributes: ['id', 'courseCode'],
+      include: {
+        model: CourseTranslation,
+        attributes: ['language', 'courseName', 'department'],
+      }
+    }) as CourseWithTranslation;
+  
+    if (!course) throw new CustomError(`course with an id ${courseId} not found`, HttpCode.NotFound);
```
no-comma, line 384: We don't want one line if statement.

server/src/middleware/errorHandler.ts
```diff
@@ -0,0 +1,70 @@
+// SPDX-FileCopyrightText: 2023 The Aalto Grades Developers
+//
+// SPDX-License-Identifier: MIT
+
+import { NextFunction, Request, Response } from 'express';
+import { ValidationError } from 'yup';
+
+export enum HttpCode {
```
no-comma, line 8: I think this enum is a great idea, but it could be moved under `types` instead and the 200 status could be included as well.

server/src/middleware/errorHandler.ts
```diff
@@ -0,0 +1,70 @@
...
+      success: false,
+      errors: err.errors
+    });
+    return;
+  }
+
+  // Fallback if no other error matches
+  res.status(HttpCode.InternalServerError).send({
+    success: false,
+    errors: ['Internal Server Error']
```
no-comma, line 67: Should be in all lowercase: `'internal server error'`.

server/src/middleware/errorHandler.ts
```diff
@@ -0,0 +1,70 @@
...
+export function ErrorHandler(err: unknown, req: Request, res: Response, next: NextFunction): void {
+
+  if (err instanceof CustomError) {
+    res.status(err.statusCode);
+
+    if (err instanceof SisuError) {
+      res.send({
+        success: false,
+        errors: [err.message],
+        sisuStatusCode: err.sisuErrorCode
```
no-comma, line 44: Should this be here at all? We're already including the Sisu status in the error message, though it could be justified to include it separately as well. We could discuss this at some point.

server/src/middleware/errorHandler.ts
```diff
@@ -0,0 +1,70 @@
...
+
+export enum HttpCode {
+  BadRequest = 400,
+  Unauthorized = 401,
+  NotFound = 404,
+  InternalServerError = 500,
+  BadGateway = 502
+}
+
+export class CustomError extends Error {
```
no-comma, line 16: `CustomError` might be too generic. I suggested `AaltoGradesError` on a whim at one point, though we could probably discuss this and collectively agree on a name.

server/src/app.ts
```diff
@@ -5,6 +5,7 @@
 import express, { Application } from 'express';
 import { router } from './routes/index';
 import cors from 'cors';
+import { ErrorHandler } from './middleware/errorHandler';
```
no-comma, line 8: There should be an empty line between this and the previous import.

server/src/controllers/course.ts
```diff
@@ -405,31 +420,13 @@ export async function getCourse(req: Request, res: Response): Promise<Response>
...
-        success: false,
-        error: error.message
-      });
-    }
-
-    return res.status(500).send({
-      success: false,
-      error: 'Internal Server Error'
-    });
+    next(error);
```
Gskartwii, line 429: Would it be reasonably possible to let this error propagate without catching it here?

server/src/middleware/errorHandler.ts
```diff
@@ -0,0 +1,70 @@
...
+export class SisuError extends CustomError {
+  public readonly sisuErrorCode: number;
+
+  constructor(sisuErrorCode: number) {
+    super(`external API error: ${sisuErrorCode}`, HttpCode.BadGateway);
+    this.sisuErrorCode = sisuErrorCode;
+  }
+}
+
+export function ErrorHandler(err: unknown, req: Request, res: Response, next: NextFunction): void {
```
Gskartwii, line 35: The `NextFunction` doesn't seem to be get used. While there probably isn't currently a middleware that could do anything after this, it would feel like good practice to call it after processing the error. Correct me if I'm wrong!

Gskartwii (changes requested):
Great work! Let's keep extending these changes to the rest of the routes.

server/src/controllers/course.ts
jj-stigell: Old habits die hard 😅

server/src/middleware/errorHandler.ts
jj-stigell: moved under types and added few common http codes

server/src/middleware/errorHandler.ts
jj-stigell: This was more to demonstrate how we can make even more customized errors from our "base" custom error. Unless client needs the sisu status code in number format this is useless. If we don't need to send the sisu status code separately, then we don't need `SisuError`. We could just use the `customError` in the sisu controllers as follows:
```javascript
throw new CustomError(`external API error: ${sisuStatusCode}`, HttpCode.BadGateway);
```

If we don't need the `SisuError`, I could delete it for now. Later on we can create new errors if we need properties/methods etc. that are not present in the base error.

What do you think?

server/src/controllers/course.ts
jj-stigell: According to express [docs](https://expressjs.com/en/guide/error-handling.html), with async functions we have to use try...catch and pass them to the next. Seems like that with express 5, there is no need for try...catch as express will automatically catch and handle errors. Too bad its still in beta 🥲

server/src/middleware/errorHandler.ts
jj-stigell: Reason why the `NextFunction` is declared here is that if not present, express will use its own default error handler. I don't know what kind of possible side-effects there is if the `NextFunction` is or is not used in the last middleware.

server/src/controllers/course.ts
Gskartwii: Ah, that's very unfortunate. Could we write a higher-order function that automatically converts an asynchronous middleware to one that passes rejections to Express through the `next` function, to avoid this boilerplate and chances of missing errors? Let's say,

```typescript
function handleErrors(func: (Request, Response, NextFunction) => Promise<void>): (Request, Response, NextFunction) => Promise<void> {
    return async (req: Request, res: Response, next: NextFunction) => {
        try {
            return func(req, res, next);
        } catch (err: unknown) {
            next(err);
        }
    };
}
```

(Pardon my TypeScript, I wrote this without using a real code editor).

server/src/middleware/errorHandler.ts
Gskartwii: Right, Express needs to make the decision based on the function arity. In that case, can we silence this lint somehow?

Gskartwii (commented):
I like the improvements that have been made. Still a bit curious about the prospect of reducing boilerplate in error handling even further.

server/src/middleware/errorHandler.ts
no-comma: Frontend developers said that they likely wouldn't need Sisu's status code, so this error type could be deleted.

server/src/middleware/errorHandler.ts
no-comma: `ApiError` was decided as the name.

server/src/middleware/errorHandler.ts
no-comma: `// eslint-disable-next-line no-unused-vars` may be one way to do it.

jj-stigell: eslint disabled 649b66633150df554734a031ce62fd34fd09718a

server/src/middleware/errorHandler.ts
jj-stigell: sisu error type removed a9dbb68d1d42dc62c64a95e5554e8c19e16c405e

server/src/controllers/course.ts
jj-stigell: Added `handleErrors` function to the routes 024fd9c1d56361742eef5483c0fb0fd428934094.

I had problems with missing properties (+80) and could not find any quick solutions for the problem except using `any` and disabling linter for those. I think maybe we need to create some kind of custom type that has all the missing properties?

server/src/controllers/courseInstance.ts
```diff
@@ -48,27 +50,106 @@ export async function getCourseInstance(req: Request, res: Response): Promise<vo
...
+  });
 
-    const responsibleTeacher: User = await findUserById(instance.responsibleTeacher);
+  res.status(HttpCode.ok).send({
+    success: true,
+    data: {
+      instance: parsedInstanceData
+    }
+  });
+  return;
```
no-comma, line 115: Unnecessary.

server/src/controllers/courseInstance.ts
```diff
@@ -84,158 +165,31 @@ export async function getCourseInstance(req: Request, res: Response): Promise<vo
...
+      courseInstances: instancesData
     }
-
-    return res.status(500).send({
-      success: false,
-      error: 'internal server error'
-    });
-  }
+  });
+  return;
```
no-comma, line 192: Unnecessary.

server/src/controllers/courseInstance.ts
```diff
@@ -291,68 +245,39 @@ export async function addCourseInstance(req: Request, res: Response): Promise<vo
...
     }
-
-    res.status(401).send({
-      success: false,
-      error: error
-    });
-    return;
-  }
+  });
+  return;
```
no-comma, line 282: Unnecessary.

no-comma (changes requested):
Looks great for the most part, I'm mostly just nitpicking.

But to clarify, did you change all controllers to match their specs in terms of status codes and error messages?

server/test/controllers/user.test.ts
```diff
@@ -65,12 +68,29 @@ describe('Test GET /v1/user/:userId/courses', () => {
...
-    const courses: CoursesOfUser = res.body.courses as CoursesOfUser;
+    const courses: CoursesOfUser = res.body.data.courses as CoursesOfUser;
     expect(courses.current.length).toBe(1);
     expect(courses.current[0].courseCode).toBe('PHYS-A1130');
     expect(courses.previous.length).toBe(1);
     expect(courses.previous[0].courseCode).toBe('TU-A1100');
   });
+
+  it('should return fail with non-existing user id', async () => {
+    const res: supertest.Response = await request.get('/v1/user/9999999/courses');
```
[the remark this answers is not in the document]
jj-stigell: merged

server/src/types/httpCode.ts
```diff
@@ -0,0 +1,14 @@
+// SPDX-FileCopyrightText: 2023 The Aalto Grades Developers
+//
+// SPDX-License-Identifier: MIT
+
+export enum HttpCode {
+  ok = 200,
+  created = 201,
```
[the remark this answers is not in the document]
jj-stigell: Well noticed 👀

server/src/middleware/errorHandler.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  constructor(message: string, statusCode: HttpCode) {
+    super(message);
+    this.statusCode = statusCode;
+    this.name = 'ApiError';
+  }
+}
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export function ErrorHandler(err: unknown, req: Request, res: Response, next: NextFunction): void {
```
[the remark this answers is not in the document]
jj-stigell: updated

server/src/controllers/utils/course.ts
```diff
@@ -0,0 +1,23 @@
...
+import Course from '../../database/models/course';
+
+import { ApiError } from '../../middleware/errorHandler';
+import { HttpCode } from '../../types/httpCode';
+
+/**
+ * Finds a course by its ID.
+ * @param {number} courseId - The ID of the course.
+ * @returns {Promise<Course>} - The found course model object.
+ * @throws {Error} - If the course is not found, it throws an error with a message
```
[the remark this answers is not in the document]
jj-stigell: correct

Gskartwii (commented):
Very thorough refactoring, I'm taken aback! I echo the comments from @no-comma. Other than those, I have no complaints.

Gskartwii:
I've gotten rid of the `any` type manually!

server/src/routes/index.ts
```diff
@@ -35,6 +35,17 @@ router.use(cookieParser());
 router.use('/api-docs', swaggerUI.serve);
 router.get('/api-docs', swaggerUI.setup(openapiSpecification));
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function handleErrors(fn: any): any {
```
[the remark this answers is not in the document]
jj-stigell: Added TODO comment and moved to middlewares. Return cannot be void without breaking the whole thing 🙃

I was thinking about the function name. Is `handleErrors` good name since its more like passing on the possible errors and the `errorHandler` is the one actually processing errors. atm its really easy to mix these to naming wise. Would name like `routeDispatcher`, `routeHandler` or smth else be better?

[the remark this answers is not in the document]
no-comma: That's true, I don't think `route` in the name is ideal though. If I were to decide on a name right now I would pick  `controllerDispatcher`, but this is something that could be discussed more later. Not for this PR though since the sprint is going to end before we'd have a good chance to do so, so just place some reminder to come back to this.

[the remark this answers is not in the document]
jj-stigell: `controllerDispatcher` sounds good, updated to that.

server/src/middleware/errorHandler.ts
```diff
@@ -0,0 +1,57 @@
...
+  // TODO: appropriate logging in case of errors
+
+  if (err instanceof ApiError) {
+    res.status(err.statusCode);
+
+    res.send({
```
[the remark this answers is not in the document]
jj-stigell: The idea was that if we have multiple Errors extending the `ApiError`, we would set the status on "top level". then we wouldn't have to set it separately on each `res.send`. This made more sense when we still had the `SisuError`.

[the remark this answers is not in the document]
no-comma: Then I think this is fine in case in the future `ApiError` is extended further.

no-comma (approved):

server/src/controllers/course.ts
```diff
@@ -23,181 +25,142 @@ export async function addCourse(req: Request, res: Response): Promise<void> {
...
+      courseData.name.sv = translation.courseName;
+      break;
 
-    res.json({
-      success: true,
-      course: courseData
-    });
-  } catch (error) {
-    // TODO: appropriate logging in case of errors
-    await t.rollback();
```
Gskartwii, line 105: I think we've lost this rollback in the refactoring process. Sequelize does seem to provide a very handy way to automatically roll back, though: https://sequelize.org/docs/v6/other-topics/transactions/#managed-transactions

Gskartwii (changes requested):
One last nitpick about Sequelize transactions

Gskartwii (approved):
Perfect!

PR #147, opened 2023-02-13
Documentation for session management routes

Gskartwii:


server/src/routes/session.ts
```diff
@@ -0,0 +1,155 @@
...
+ *     responses:
+ *       200:
+ *         description: The user has been created and logged in
+ *         content:
+ *           application/json:
+ *             schema:
+ *               $req: '#/definitions/SignupResult'
+ *         headers:
+ *           Set-Cookie:
+ *             description: 'JWT token, storing an access token to allow using the API'
```
no-comma, line 113: Unnecessary quotation marks?

server/src/routes/session.ts
```diff
@@ -0,0 +1,155 @@
...
+ *       logging in to the system.
+ *     requestBody:
+ *       description: Basic user information
+ *       content:
+ *         application/json:
+ *           schema:
+ *             $ref: '#/definitions/SignupRequest'
+ *     responses:
+ *       200:
+ *         description: The user has been created and logged in
```
no-comma, line 106: This isn't the only place this applies, but I think all descriptions should end with proper punctuation, to be consistent with the very likely possibility of multiple sentence descriptions.

server/src/routes/session.ts
```diff
@@ -0,0 +1,155 @@
...
+ *       200:
+ *         description: The user has successfully logged out
+ *         content:
+ *           application/json:
+ *             schema:
+ *               type: object
+ *               properties:
+ *                 success:
+ *                   type: boolean
+ *                   description: Whether the operationsucceeded
```
no-comma, line 73: Typo.

server/src/routes/session.ts
```diff
@@ -0,0 +1,155 @@
...
+ */
+router.post('/v1/auth/login', express.json(), authLogin);
+
+/**
+ * @swagger
+ * /v1/auth/logout:
+ *   post:
+ *     tags: [Session]
+ *     description: >
+ *       Log out of a session, removing the session key from the browser.
```
no-comma, line 54: Isn't removing the session key from the browser the job of the client?

server/src/routes/session.ts
```diff
@@ -0,0 +1,155 @@
...
+ *     responses:
+ *       200:
+ *         description: The login has succeeded, and an access token is return in a cookie.
+ *         content:
+ *           application/json:
+ *             schema:
+ *               $ref: '#/definitions/LoginResult'
+ *         headers:
+ *            Set-Cookie:
+ *              description: 'JWT token, storing an access token to allow using the API'
```
no-comma, line 33: Unnecessary quotation marks again?

server/src/routes/index.ts
```diff
@@ -101,6 +118,91 @@ router.get('/api-docs', swaggerUI.setup(openapiSpecification));
  *             description: Previous Courses
  *             items:
  *               $ref: '#/definitions/Course'
+ *   Credentials:
```
no-comma, line 121: If Swagger allows for it, these definitions should be made in the file that uses them, so in this case `routes/session.ts`.

no-comma (changes requested):

server/src/routes/auth.ts
```diff
@@ -0,0 +1,238 @@
...
+ *             type: number
+ *             description: >
+ *               The database identifier of the user
+*/
+
+/**
+ * @swagger
+ * /v1/auth/login:
+ *   post:
+ *     tags: [Session]
```
no-comma, line 100: Perhaps `Auth` as tag instead?

server/src/routes/index.ts
```diff
@@ -39,7 +39,24 @@ router.get('/api-docs', swaggerUI.setup(openapiSpecification));
...
  * definitions:
+ *   Failure:
+ *     type: object
+ *     description: A reason for a failure, with a chiefly developer-facing error message
+ *     properties:
+ *       success:
+ *         type: boolean
+ *         description: '`false` to indicate failure'
+ *         example: false
+ *       message:
```
no-comma, line 57: If `Failure` is representing what gets returned upon failure from any API route, then this should be `errors`, which is an array of all errors which occurred, since Yup may return multiple errors.

server/src/routes/index.ts
```diff
@@ -39,7 +39,24 @@ router.get('/api-docs', swaggerUI.setup(openapiSpecification));
...
+ *       in: cookie
+ *       name: jwt
  * definitions:
+ *   Failure:
+ *     type: object
+ *     description: A reason for a failure, with a chiefly developer-facing error message
+ *     properties:
+ *       success:
+ *         type: boolean
+ *         description: '`false` to indicate failure'
```
no-comma, line 55: Quotation marks.

no-comma (changes requested):
Just a couple more things.

server/src/routes/index.ts
Gskartwii: These quotation marks turned out to be necessary, as otherwise YAML doesn't allow string literals that start with a backtick.

no-comma (approved):

jj-stigell (approved):

PR #148, opened 2023-02-15
Add nonblock statement ESLint rule

no-comma:
Adds an ESLint rule to enforce avoiding (primarily) one-line if statements.
```ts
if (foo) bar();
else throw new Error();
```
will be corrected to:
```ts
if (foo)
  bar();
else
  throw new Error();
```
I would have included this in the refactoring PR but I only discovered this rule now.

### DIFF
--- a/server/.eslintrc.json
+++ b/server/.eslintrc.json
@@ -51,6 +51,10 @@
       "error",
       "always"
     ],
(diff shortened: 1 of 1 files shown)


Gskartwii (approved):

jj-stigell (approved):
👍

PR #150, opened 2023-02-15
Course instance refactoring and documentation

no-comma:


server/src/controllers/utils/course.ts
```diff
@@ -10,14 +10,15 @@ import { HttpCode } from '../../types/httpCode';
 /**
  * Finds a course by its ID.
  * @param {number} courseId - The ID of the course.
+ * @param {HttpCode} errorCode - HTTP status code to return if the course was not found.
  * @returns {Promise<Course>} - The found course model object.
  * @throws {ApiError} - If the course is not found, it throws an error with a message
  * indicating the missing course with the specific ID.
  */
-export async function findCourseById(courseId: number): Promise<Course> {
+export async function findCourseById(courseId: number, errorCode: HttpCode): Promise<Course> {
```
Gskartwii, line 18: It could be preferable not to have this function be concerned with the express.js API, even if if means handling the error explicitly outside this function. How about returning `Promise<Course | null>`, and letting the caller decide how to handle the situation?

server/src/controllers/utils/user.ts
```diff
@@ -10,14 +10,16 @@ import { HttpCode } from '../../types/httpCode';
 /**
  * Finds a user by its ID.
  * @param {number} userId - The ID of the user to be found.
+ * @param {HttpCode} errorCode - HTTP status code to return if the user was not found.
  * @returns {Promise<User>} - A promise that resolves with the found user model object.
  * @throws {ApiError} - If the user is not found, it throws an error with a message
  * indicating the missing user with the specific ID.
  */
-export async function findUserById(userId: number): Promise<User> {
+export async function findUserById(userId: number, errorCode: HttpCode): Promise<User> {
```
Gskartwii, line 18: Same applies here!

Gskartwii (commented):
Impressively speedy refactoring. I've made a small suggestion about separation of concerns, which may or may not be useful.

server/src/controllers/utils/course.ts
no-comma: If these functions don't throw errors then I don't think it makes sense to have them at all since they would just be simple wrappers for very basic queries.

Since these are helper functions specifically for use with controllers, I don't see the problem with having them throw errors. If there is a scenario in which we want to find a course and not throw an error if it isn't found, we can simply do that with `Course.findByPk` directly.

jj-stigell: What are the possible cases when we would not want to throw the error inside this function? I could imagine some situations like when user or course information is optional. In these situations would it be better to not use these util functions at all and just query with the models?

Gskartwii: I will have to agree that since these are not services but they are controllers, it's okay to have the HTTP status code here, after all.

Gskartwii (approved):

jj-stigell (approved):

PR #152, opened 2023-02-16
Create assignment view

viivi-uhari:
Created a view for assignment creation. Not connected to backend yet and formula selection shouldn't be working either.


client/src/services/assignments.js
```diff
@@ -0,0 +1,80 @@
+// SPDX-FileCopyrightText: 2023 The Aalto Grades Developers
+
+// SPDX-License-Identifier: MIT
+
+// The parameter 'indices' used in the following functions is an array of integres
```
marillam, line 5: You've made the indices work so the code here seems good to me. If any problems occur with the mapping to parent assignments, would consider ids or something. But no immediate changes needed imo.

client/src/services/assignments.js
```diff
@@ -0,0 +1,80 @@
...
+};
+
+// Set the proprety of the object that is in the location specified by the indices in the assignments array,
+// the property is set to have the value given as a parameter
+const setProperty = (indices, assignments, property, value) => {
+  const lastIndex = indices[indices.length - 1];
+  const indicesWithoutLast = indices.slice(0, -1);
+  const array = indicesWithoutLast.reduce((acc, current_index) => acc[current_index].subAssignments, assignments);
+  array[lastIndex][property] = value;
+};
```
marillam, line 78: Maybe move these two above the others, so they've been defined before the function using them.

marillam (commented):
In general, everything looks good to me. And this isn't connected to backend so new changes there shouldn't be an issue here.

client/src/services/assignments.js
viivi-uhari: Good idea, I'll probably look into this while implementing the edit assignment view and see if this will be an easy change to make.

marillam (approved):
Amazing, great work

client/src/components/CreateAssignmentView.js
```diff
@@ -0,0 +1,77 @@
...
+//
+// SPDX-License-Identifier: MIT
+
+import React, { useState }  from 'react';
+import Typography from '@mui/material/Typography';
+import Container from '@mui/material/Container';
+import Button from '@mui/material/Button';
+import Box from '@mui/material/Box';
+import Assignment from './create-assignment/Assignment';
+import subAssignmentServices from '../services/assignments';
```
kiiamakir, line 11: Could this be imported just as assignmentServices? Since it also contains a service for removing an assignment etc. Same with the other files

client/src/components/create-assignment/LeafAssignment.js
```diff
@@ -0,0 +1,177 @@
...
+      updatedAssignments = subAssignmentServices.setProperty(indices, updatedAssignments, 'name', value);
+    }
+    setAssignments(updatedAssignments);
+  };
+
+  const getValue = (fieldData) => {
+    if (fieldData.fieldId === 'category') {
+      return subAssignmentServices.getProperty(indices, assignments, 'category');
+    } else {
+      return console.log(fieldData.fieldId);
```
kiiamakir, line 63: No return needed here

kiiamakir (changes requested):
Looks good! Just a couple small changes

kiiamakir (approved):
All good!

Issue #138, opened 2023-02-01
Attainment APIs


no-comma:
As a frontend developer, I want to have an API to add  and edit attainment attributes (for example weight), so they can be stored through the UI

- [x] API to add attainment @jj-stigell 
- [x] Schema change for attainment, max points, weight, min requirement @jj-stigell 
- [x] API to edit attainment @jj-stigell 
- [x] API to get a single attainment, optionally including its children or descendants @Tonits @jyrge  
- [ ] #213 
- [x] API to delete an attainment and all of its descendants @no-comma 
- [x] Decide on API structure

no-comma:
This issue only has one task left, which has its own issue, so I think this is safe to close now.

PR #153, opened 2023-02-17
APIs for adding and editing attainments

jj-stigell:
#138 

Adds functionality to add and edit assignments.
Has following attributes atm:
- id (PK, assigned automatically on creation)
- course instance id (FK)
- name
- execution date
- expiry date

When editing assignment all except id is optional.


server/src/routes/assignment.ts
```diff
@@ -0,0 +1,126 @@
...
+ *               $ref: '#/definitions/Failure'
+ *       404:
+ *         description: Course instance does not exist.
+ *         content:
+ *           application/json:
+ *             schema:
+ *               $ref: '#/definitions/Failure'
+ */
+router.post(
+  '/v1/assignment',
```
no-comma, line 77: Since course assignments are instance specific, this route could maybe be `/v1/courses/instances/:instanceId/assignments` and the instance ID can be removed from the request body to be consistent with other routes. Assignments could also be a candidate for using composite keys in the next sprint.

I guess this is fine for now though since we don't really have time to collectively decide on specifications for these routes at the moment.

server/src/routes/index.ts
```diff
@@ -33,6 +34,7 @@ export const router: Router = Router();
 router.use(cookieParser());
 router.use(authRouter);
 router.use(userRouter);
+router.use(assignmentRouter);
```
no-comma, line 37: I would order these alphabetically.

server/src/database/models/courseAssignment.ts
```diff
@@ -7,16 +7,17 @@ import {
...
   InferAttributes<CourseAssignment>,
   InferCreationAttributes<CourseAssignment>
 > {
   declare id: CreationOptional<number>;
-  declare courseInstancePartialGradeId: ForeignKey<CourseInstancePartialGrade['id']>;
-  declare assignmentId: string;
-  declare maxPoints: number;
+  declare courseInstanceId: ForeignKey<CourseInstance['id']>;
+  declare name: string;
+  declare executionDate: Date;
```
no-comma, line 19: This name is a bit confusing, what exactly does this attribute represent?

server/src/controllers/utils/courseInstance.ts
```diff
@@ -0,0 +1,23 @@
...
+ * Finds a course instance by its ID.
+ * @param {number} id - The ID of the course instance.
+ * @returns {Promise<CourseInstance>} - The found course instance model object.
+ * @throws {ApiError} - If the course instance is not found, it throws an error with a message
+ * indicating the missing course instance with the specific ID.
+ */
+export async function findCourseInstanceById(id: number): Promise<CourseInstance> {
+  const courseInstance: CourseInstance | null = await CourseInstance.findByPk(id);
+  if (!courseInstance) {
+    throw new ApiError(`course instance with ID ${id} not found`, HttpCode.NotFound);
```
no-comma, line 20: This should also take the error code as a parameter, since we may not necessarily always want to return 404 when a course instance isn't found.

server/src/controllers/utils/courseAssignment.ts
```diff
@@ -0,0 +1,23 @@
...
+ * Finds a course assignment by its ID.
+ * @param {number} id - The ID of the course assignment.
+ * @returns {Promise<CourseAssignment>} - The found course assignment model object.
+ * @throws {ApiError} - If the course assignment is not found, it throws an error
+ * with a message indicating the missing course assignment with the specific ID.
+ */
+export async function findCourseAssignmentById(id: number): Promise<CourseAssignment> {
+  const courseAssignment: CourseAssignment | null = await CourseAssignment.findByPk(id);
+  if (!courseAssignment) {
+    throw new ApiError(`assignment with ID ${id} not found`, HttpCode.NotFound);
```
no-comma, line 20: Same comment about error codes here.

server/src/routes/assignment.ts
```diff
@@ -0,0 +1,126 @@
...
+ *             schema:
+ *               $ref: '#/definitions/Assignment'
+ *       400:
+ *         description: Creation failed, due to malformed/missing parameters.
+ *         content:
+ *           application/json:
+ *             schema:
+ *               $ref: '#/definitions/Failure'
+ *       404:
+ *         description: Course instance does not exist.
```
no-comma, line 109: 422 might be a better code in this case, at least when the instance ID is passed in the body. But this again is something that should be discussed later.

server/src/routes/assignment.ts
```diff
@@ -0,0 +1,126 @@
...
+);
+
+/**
+ * @swagger
+ * /v1/assignment/{assignmentId}:
+ *   put:
+ *     tags: [Assignment]
+ *     description: Update existing assignment.
+ *     requestBody:
+ *       description:  Assignment data to be updated.
```
no-comma, line 90: Extra space.

server/src/routes/assignment.ts
```diff
@@ -0,0 +1,126 @@
...
+ *           application/json:
+ *             schema:
+ *               $ref: '#/definitions/Failure'
+ *       404:
+ *         description: Course instance does not exist.
+ *         content:
+ *           application/json:
+ *             schema:
+ *               $ref: '#/definitions/Failure'
+ *       404:
```
no-comma, line 114: I'm not sure if it's good practice to put two of the same status code in this list. I would only have one 404 with a description saying something like: `The given course instance or assignment does not exist.`

server/src/types/course.ts
```diff
@@ -48,3 +48,11 @@ export interface CourseInstanceData {
   responsibleTeacher?: string | undefined,
   responsibleTeachers?: Array<string>,
 }
+
+export interface Assignment {
```
no-comma, line 52: Maybe `CourseAssignmentData`?

server/src/controllers/courseAssignment.ts
```diff
@@ -0,0 +1,98 @@
...
+    executionDate: yup
+      .date()
+      .notRequired(),
+    expiryDate: yup
+      .date()
+      .notRequired()
+  });
+
+  const id: number = Number(req.params.assignmentId);
+  await requestSchema.validate({ id: id, ...req.body }, { abortEarly: false });
```
no-comma, line 75: I think it would be better to validate the ID from the URL separately with `idSchema`.

no-comma (changes requested):

server/src/routes/assignment.ts
no-comma: Thinking about this now I don't really like the half measure of putting `/v1/courses/instances` at the start of this route without the course ID, since we would also have routes which do have it like `/v1/courses/:courseId/instances`, which seems messy to me.

My worry with using composite keys for assignments is that we would end up with routes like `/v1/courses/:courseId/instances/:instanceId/assignments/:assignmentId` which might be too complex for its own good.

But on the other hand if we want to get all assignments of a course instance, without composite keys we would still somehow need to include the course instance ID in the URL, like `/v1/course-instances/:instanceId/assignments`, but then we run into the problem of inconsistency like with the `/v1/courses/:courseId/instances` and `/v1/instances` problem we discussed before.

You can ignore this comment for this PR, but we should consider this later.

server/src/database/models/courseAssignment.ts
jj-stigell: Added comment for clarification

viivi-uhari (commented):
Looks good but the an assignment should have a parent ID attribute to its parent assignment. If an assignment doesn’t have an assignment as its parent then the parent is the course instance. Also, a simple label/tag should be generated for each assignment to help the work of a teacher (for example a tag of six characters: `'KLMX5R'`). And finally, a route that gets all the assignments of an instance should also be implemented, but these additions can be done in another PR.

I also think that a route that contains the instance id (something like `/v1/instances/:instanceId/assignments/:assignmentId`) would be useful but I don’t have strong opinion about this and I guess the APIs need to be discussed by the backend team either way.

Also as a note, a route that gets an assignment based on its ID and all of its sub assignments (even indirect sub assignments) would be useful. I have implemented a simple function on the client side that gets all of the children of an assignment (from all assignments) and returns a tree structure, but I think this is something that would be better to do on the server if possible. But again this could be discussed by the backend team that what would be the best approach: client or server (or DB)?

jj-stigell:
Adding and updating assignments update based on API and terms meetings.

All client facing references to the assignment now use the word "attainment". Also API documentation uses the word "attainment". Backend code and database use the word "attainable". If you have strong opinions/think these are named incorrectly, please feel free to make the changes by yourself. From here on in this message I will refer to assignments as attainments.

If the attainments is a child of some other attainment, the parent attainments id is saved in the `attainableId` field in the database. Idea was to use attribute name `parentId` but for some reason when querying sequelize gives warning that `attainableId` not found in the table `attainable`.

It seems that sequelize forces the FK name as `model + key` which is in this case `attainableId` even if the model and db attribute are explicitly named as `parentId` in the model and migrations. I will investigate this when more time at hand. For now added TODO memo in the attainment model. If anyone have fix for this, please feel free to fix this issue in this PR. **NOTE** at the moment API uses the `parentId` so updating the backend naming does not brake the API.

During API design meetings, there was no discussion on what kind of error message and http codes we send in the following situations:
- instance does not belong to the course. atm `instance ID {courseInstanceId} does not belong to the course ID {courseId}` http code: 409 conflict
- parent attainment does not belong to the instance where new attainment is being created. atm `parent attainment ID {parentId} does not belong to the instance ID {courseInstanceId}` http code: 409 conflict

Are these good or changes required?

**One question concerning the attainment updating.** Do we also want to have the option to update the course, instance and parent id of the attainment? Atm only the non-key attributes (`name`, `expiryDate`, and `executionDate`) are updatable.

viivi-uhari (commented):
Good changes! But I have some questions and notes.

Is it expected here that the client returns only one assignment with a parent ID to the server when creating (or editing) an assignment? Wasn’t the idea that the client could return a tree of assignments? Or am I just understanding the code incorrectly?

Also, how will the 'tag' (generated from the course id, instance id, and attainment id) be created and returned to the client?

Finally, when a user edits an assignments, it is possible to delete its sub-assignments as well as sub-assignments under those. It is also possible to delete the whole top assignment all together. So, because of these use cases, should the parentId be modifiable?

jj-stigell:
> Is it expected here that the client returns only one assignment with a parent ID to the server when creating (or editing) an assignment? Wasn’t the idea that the client could return a tree of assignments? Or am I just understanding the code incorrectly?

You are reading the code correctly. Right now assignments are added/edited one by one (both parent and child assignments). There is a chance that I might have misunderstood the way we implement this. Was the idea that client sends a batch of (possibly nested) assignments to the server for adding/updating? In this case, how do you want the assignment ids and tags to be returned, also nested?

> Also, how will the 'tag' (generated from the course id, instance id, and attainment id) be created and returned to the client?

My bad, totally forgot this one. I will implement one that returns also the tag with the updated/added attainment.

> Finally, when a user edits an assignments, it is possible to delete its sub-assignments as well as sub-assignments under those. It is also possible to delete the whole top assignment all together. So, because of these use cases, should the parentId be modifiable?

At the moment, the edit endpoint does not allow deletion. I would keep the updating and deleting resources as separate endpoints. I can update the `parentId` to be modifiable 👍. How we delete the sub-assignments should be discussed more, maybe we could use url params like `DELETE ...assignments/:assignmentId?deleteChildren=(boolean)` OSLT.

server/src/database/models/attainable.ts
```diff
@@ -10,24 +10,23 @@ import { sequelize } from '..';
...
-  declare type: string;
-  declare platform: string;
-  declare maxPoints: number;
-  declare minPoints: number;
-  declare weight: number;
-  declare expireAt: Date;
+  // TODO rename to parentId, atm sequelize forces name as "model + key" when querying.
+  declare attainableId: CreationOptional<ForeignKey<Attainable['id']>>;
+  declare name: string;
+  declare executionDate: Date; // Date when assignment is done (e.g., deadline or exam date)
```
no-comma, line 23: Perhaps `dueDate` then?

server/src/database/models/index.ts
```diff
@@ -2,16 +2,24 @@
...
-import CourseInstancePartialGrade from './courseInstancePartialGrade';
 import CourseResult from './courseResult';
 import CourseRole from './courseRole';
 import CourseTranslation from './courseTranslation';
 import User from './user';
-import UserAssignmentGrade from './userAssignmentGrade';
-import UserPartialGrade from './userPartialGrade';
+import UserAttainableGrade from './userAttainableGrade';
+
+Attainable.hasMany(Attainable, {
```
no-comma, line 14: This is more of a general comment than this PR, but why are all the associations defined in this one file? Wouldn't it be better to define, e.g., `Course` associations in `models/course.ts`?

server/src/database/models/userAttainableGrade.ts
```diff
@@ -7,22 +7,22 @@ import {
...
 
 import { sequelize } from '..';
-import CourseAssignment from './courseAssignment';
+import Attainable from './attainable';
 import User from './user';
 
-export default class UserAssignmentGrade extends Model<
-  InferAttributes<UserAssignmentGrade>,
-  InferCreationAttributes<UserAssignmentGrade>
+export default class UserAttainableGrade extends Model<
```
no-comma, line 13: "Study attainment" was chosen as the term to replace "assignment," but the reason we thought of using the name `Attainable` instead of `Attainment` in the backend to represent what we previously called "assignments" is because "attainment" gives an impression that something has already been attained. But in context, the "study attainments" aren't something that have been attained, they are something to be attained, so the term "attainable" was suggested.

But this model does represent something that has already been attained, so `AttainmentGrade` would be an appropriate name, or maybe `AttainedGrade`.

server/src/routes/attainable.ts
```diff
@@ -0,0 +1,140 @@
...
+ *         required: true
+ *         description: Study attainment name.
+ *       executionDate:
+ *         type: string
+ *         required: true
+ *         description: Date when the attainment is completed (e.g. deadline date or exam date).
+ *       expiryDate:
+ *         type: string
+ *         required: true
+ *         description: Date when the attainment expires.
```
no-comma, line 35: This should also explain what it means for an attainment to expire. Something like:

> Once an attainment has expired, it is no longer eligible to count as completion for future course instances.

server/src/routes/attainable.ts
```diff
@@ -0,0 +1,140 @@
...
+ *             schema:
+ *               $ref: '#/definitions/Attainment'
+ *       400:
+ *         description: Creation failed, due to validation errors or missing parameters.
+ *         content:
+ *           application/json:
+ *             schema:
+ *               $ref: '#/definitions/Failure'
+ *       404:
+ *         description: Course, course instance, or parent study attainment does not exist.
```
no-comma, line 82: I don't think `404` is appropriate if the parent attainment doesn't exist, since it's not included in the URL.

I think I already mentioned this in a previous review, but we've used `422` in `POST /v1/courses/:courseId/instances` when the ID of the given teacher in charge is invalid.

https://developer.mozilla.org/en-US/docs/web/http/status/422

> The HyperText Transfer Protocol (HTTP) 422 Unprocessable Content response status code indicates that the server understands the content type of the request entity, and the syntax of the request entity is correct, but it was unable to process the contained instructions.

viivi-uhari:
Okay great, thanks for the clarifications! 

And yes the idea is that a tree (nested) structure would be returned from the client and a tree structure should also be returned from the server. This is because when creating and editing attainments, new (sub-) attainments can be created and all of the attainments will be returned to the server at the same time. But we don't want to create IDs or parent IDs on the client side so it's not possible to return them as a list without loosing the structure of them. Here's an example what I was thinking:

**Attainment creation**
Client structure that is returned to server:
```
{ name: '2 optional exercises',
  date: new Date(2023, 9, 1), 
  expiryDate: new Date(2024, 8, 14), 
  subAssignments: [
    { name: 'Exercise 1',
      date: new Date(2023, 9, 1), 
      expiryDate: new Date(2024, 8, 14),
      subAssignments: [],
    }, 
    { name: 'Exercise 2',
      date: new Date(2023, 9, 1), 
      expiryDate: new Date(2024, 8, 14),
      subAssignments: [],
    }
  ] 
}
```

What it should be modified to on the server side:
```
[{ id: 1,
    parentId: null,  // course instance is the parent
    name: '2 optional exercises'
    date: new Date(2023, 9, 1), 
    expiryDate: new Date(2024, 9, 1), 
 },
 { id: 2,
    parentId: 1,
    name: 'Exercise 1',
    date: new Date(2023, 9, 1), 
    expiryDate: new Date(2024, 9, 1),  
  },
  { id: 3,
    parentId: 1,
    name: 'Exercise 2',
    date: new Date(2023, 9, 1), 
    expiryDate: new Date(2024, 9, 1), 
 }]
```

**Attainment editing**
Server converts the attainments from the DB again to a tree structure like this that is returned to the client (the IDs and parent IDs can and probably should be returned):
```
{ id: 1,
  parentId: null,
  name: '2 optional exercises',
  date: new Date(2023, 9, 1), 
  expiryDate: new Date(2024, 8, 14), 
  subAssignments: [
    { id: 2,
      parentId: 1,
      name: 'Exercise 1',
      date: new Date(2023, 9, 1), 
      expiryDate: new Date(2024, 8, 14),
      subAssignments: [],
    }, 
    { id: 3,
      parentId: 1,
      name: 'Exercise 2',
      date: new Date(2023, 9, 1), 
      expiryDate: new Date(2024, 8, 14),
      subAssignments: [],
    }
  ] 
}

```
Client adds a new sub-attainment to the tree and returns this to the server (observe that now again the IDs for the new assignment are missing and should be created on the server side):
```
{ id: 1,
  parentId: null,
  name: '2 optional exercises',
  date: new Date(2023, 9, 1), 
  expiryDate: new Date(2024, 8, 14), 
  subAssignments: [
    { id: 2,
      parentId: 1,
      name: 'Exercise 1',
      date: new Date(2023, 9, 1), 
      expiryDate: new Date(2024, 8, 14),
      subAssignments: [],
    }, 
    { id: 3,
      parentId: 1,
      name: 'Exercise 2',
      date: new Date(2023, 9, 1), 
      expiryDate: new Date(2024, 8, 14),
      subAssignments: [],
    }, 
   {  name: 'Exercise 3',
      date: new Date(2023, 9, 1), 
      expiryDate: new Date(2024, 8, 14),
      subAssignments: [],
    }, 
  ] 
}
```

So, this is what we thought with the frontend team that would be the best approach but if this seems too complex we could have another meeting about this. However, after a lot of thinking we haven’t been able to come up with a better solution.

server/test/controllers/attainable.test.ts
```diff
@@ -0,0 +1,293 @@
+// SPDX-FileCopyrightText: 2023 The Aalto Grades Developers
+//
+// SPDX-License-Identifier: MIT
+
+import supertest from 'supertest';
+
+import { app } from '../../src/app';
```
no-comma, line 7: I think this counts as an "other" import in our [code style](https://github.com/aalto-grades/base-repository/wiki/Code-style#import-order) and should be grouped with the type imports below.

server/test/controllers/attainable.test.ts
```diff
@@ -0,0 +1,293 @@
...
+
+const payload: AttainableData = {
+  name: 'exam assignment 1.1',
+  executionDate: new Date('2022-02-02'),
+  expiryDate: new Date('2022-01-01')
+};
+
+describe('Test POST /v1/courses/:courseId/instances/:instanceId/attainments', () => {
+
+  it('should create a new assignment when course and course instance exists', async () => {
```
no-comma, line 24: Attainment / attainable.

server/test/controllers/attainable.test.ts
```diff
@@ -0,0 +1,293 @@
...
+      .send(payload)
+      .set('Content-Type', 'application/json');
+
+    expect(res.body.success).toBe(true);
+    expect(res.body.errors).not.toBeDefined();
+    expect(res.body.data.attainment.id).toBeDefined();
+    expect(res.statusCode).toBe(HttpCode.Ok);
+  });
+
+  it('should create a new assignment with parent attainment', async () => {
```
no-comma, line 36: Attainment / attainable.

server/test/controllers/attainable.test.ts
```diff
@@ -0,0 +1,293 @@
...
+        `parent attainment ID ${id} does not belong to the instance ID 2`);
+      expect(res.statusCode).toBe(HttpCode.Conflict);
+    });
+});
+
+describe('Test PUT /v1/courses/:courseId/instances/:instanceId/attainments/:attainmentId', () => {
+  let id: number;
+
+  it('should update field succesfully on an existing attainment', async () => {
+    // Create a new assignment.
```
no-comma, line 207: Attainment / attainable.

server/src/controllers/attainable.ts
```diff
@@ -0,0 +1,120 @@
...
+import CourseInstance from '../database/models/courseInstance';
+
+import { ApiError } from '../types/error';
+import { idSchema } from '../types/general';
+import { HttpCode } from '../types/httpCode';
+import { findAttainableById } from './utils/attainable';
+import { findCourseById } from './utils/course';
+import { findCourseInstanceById } from './utils/courseInstance';
+
+const requestSchema: yup.AnyObjectSchema = yup.object().shape({
```
no-comma, line 20: I was going to suggest defining this in the controller below, but this schema is actually used by multiple controllers. In that case, since this is global it should be renamed to be more explicit like `addEditAttainableRequestSchema` (bad name, just an example).

no-comma (changes requested):
Sorry for the wait, now I finally had the opportunity to review this.

no-comma:
> - instance does not belong to the course. atm `instance ID {courseInstanceId} does not belong to the course ID {courseId}` http code: 409 conflict

With the current database in this PR, 409 works for this scenario.

> Also, how will the 'tag' (generated from the course id, instance id, and attainment id) be created and returned to the client?

For the implementation, we could have a util function which takes in the appropriate IDs and then computes the tag and returns it. I don't think there's any need to store it anywhere.

server/src/database/models/attainable.ts
jj-stigell: Or we could also follow the same naming convention as sequelize and use `dueAt`. Reason why `executionDate` is used is because frontend also uses this and this way we have more harmonized environment. Any comments from @viivi-uhari ?

server/src/database/models/index.ts
jj-stigell: I think the original idea was to clearly separate the models and sequelize associations. I think this worked well when we had few(er) associations, you could easily read and edit associations in one file. But now that the `index.ts` file is slowly growing bigger, maybe it would be better to have a look on this issue in the near future.

server/src/database/models/userAttainableGrade.ts
jj-stigell: renamed the model to `UserAttainmentGrade`, kept user in the name to emphasize this is user related.

jj-stigell:
> **Attainment creation** Client structure that is returned to server:
> 
> ```
> { name: '2 optional exercises',
>   date: new Date(2023, 9, 1), 
>   expiryDate: new Date(2024, 8, 14), 
>   subAssignments: [

I suppose the `date` here is actually the `executionDate`?

Add attainment route now takes in now tree structure of attainment. Controller function goes through these recursively so the depth of the sub-attainments is not limited. Each attainment is added separately to the database, possible sub-attainments are recursively added under the parent attainment. Performance wise this is not very optimal if there are hundreds of attainments added at once, but then again this endpoint is used maybe few times during the lifetime of a course instance so performance is not our top priority? Maybe we could utilize sequelize `bulkCreate` somewhere if performance becomes an issue?

Now for example this client request (to course id 1, instance id 1) for adding attainments:
```javascript
{
  name: '2 optional exercises',
  executionDate: new Date(2023, 9, 1),
  expiryDate: new Date(2024, 8, 14),
  subAssignments: [
    { name: 'Exercise 1',
      executionDate: new Date(2023, 9, 1),
      expiryDate: new Date(2024, 8, 14),
      subAssignments: [],
    },
    { name: 'Exercise 2',
      executionDate: new Date(2023, 9, 1),
      expiryDate: new Date(2024, 8, 14),
      subAssignments: [
        {
          name: 'Exercise 2.1',
          executionDate: new Date(2023, 9, 1),
          expiryDate: new Date(2024, 8, 14),
          subAssignments: [],
        }
      ],
    }
  ]
}
```

Would result in this (succesful) response from the server:
```JSON
"success": true,
"data": {
    "attainment": {
        "id": 1,
        "courseId": 1,
        "courseInstanceId": 1,
        "name": "2 optional exercises",
        "executionDate": "2023-09-01T17:00:00.000Z",
        "expiryDate": "2024-08-14T17:00:00.000Z",
        "parentId": null,
        "tag": "C1I1A1",
        "subAssignments": [
          {
            "id": 2,
            "courseId": 1,
            "courseInstanceId": 1,
            "name": "Exercise 1",
            "executionDate": "2023-09-01T17:00:00.000Z",
            "expiryDate": "2024-08-14T17:00:00.000Z",
            "parentId": 1,
            "tag": "C1I1A2",
            "subAssignments": []
          },
          {
            "id": 3,
            "courseId": 1,
            "courseInstanceId": 1,
            "name": "Exercise 2",
            "executionDate": "2023-09-01T17:00:00.000Z",
            "expiryDate": "2024-08-14T17:00:00.000Z",
            "parentId": 1,
            "tag": "C1I1A3",
            "subAssignments": [
              {
                "id": 4,
                "courseId": 1,
                "courseInstanceId": 1,
                "name": "Exercise 2.1",
                "executionDate": "2023-09-01T17:00:00.000Z",
                "expiryDate": "2024-08-14T17:00:00.000Z",
                "parentId": 3,
                "tag": "C1I1A4"
              }
            ]
          }
        ]
      }
    }
```

If you want to add new sub-attainment to existing attainment, just send the new attainment and include `parentId`. The endpoint will only return attainments (with new id's etc.) that are sent to it. If you add new sub-attainment to a existing attainment, the endpoint will not return the parent, just the newly created resources.

Edit assignment endpoint works same as previously. You can now change the `parentId` when editing attainments (Note. still has to be in the same course instance).

server/src/controllers/attainable.ts
```diff
@@ -0,0 +1,226 @@
...
+  const course: Course = await findCourseById(courseId, HttpCode.NotFound);
+
+  const instance: CourseInstance = await findCourseInstanceById(
+    courseInstanceId, HttpCode.NotFound
+  );
+
+  // Check that instance belongs to the course.
+  if (instance.courseId !== course.id) {
+    throw new ApiError(
+      `instance ID ${courseInstanceId} does not belong to the course ID ${courseId}`,
```
no-comma, line 56: I would write as: `course instance with ID [...] to the course with ID [...]`

server/src/types/course.ts
```diff
@@ -48,3 +48,17 @@ export interface CourseInstanceData {
...
+  courseId?: number,
+  courseInstanceId?: number,
+  parentId?: number,
+  tag?: string,
+  name: string,
+  executionDate: Date,
+  expiryDate: Date,
+  updatedAt?: Date,
+  createdAt?: Date,
+  subAssignments: Array<AttainableData>
```
no-comma, line 63: `subAttainables`

server/test/mockData/attainable.ts
```diff
@@ -0,0 +1,78 @@
...
+// SPDX-License-Identifier: MIT
+
+import { AttainableData } from '../../src/types/course';
+
+export const mockAttainables: AttainableData = {
+  name: '2 optional exercises',
+  executionDate: new Date(2023, 9, 1),
+  expiryDate: new Date(2024, 8, 14),
+  subAssignments: [
+    { name: 'Exercise 1',
```
no-comma, line 12: The client code does use this style, but it would be more consistent with our server code to put a newline after the opening bracket here:
```ts
subAssignments: [
  {
    name: 'Exercise 1',
    ...
  },
]
```

no-comma (changes requested):
Looks good, once you fix the tests and some minor naming details I think this can be merged.

My previous comment about changing the term `executionDate` still applies as well.

server/src/database/models/attainable.ts
viivi-uhari: I don't have string opinions about this but after thinking about this I think that `date` could actually be the best and most universal option. Since an exam isn't 'due' or assignments aren't executed in one day. Thoughts @no-comma ?

no-comma: `date` works if there is clear documentation about what it means.

viivi-uhari (commented):
Looks great now! The tree structure looks correct! And I think that the performance of this code is sufficient for now since attainments will be created/edited probably at most 50 pieces at a time. We have also been thinking about this with the frontend team that should this be divided into smaller portions since there's some performance issues on our side. But I don't think this will require changes on the server side.

Also, could the `executionDate` be changed to `date`?

And another question: Is it possible to add sub-attainments with sub-attainments while editing an attainment? So, that the higher attainment knows its parentId but the lower one doesn't because its parent is just being created. So that something like this would be returned (only the created new sub-attainments to an existing attainment with the ID 2):

```
{
  name: '2 optional exercises',
  executionDate: new Date(2023, 9, 1),
  expiryDate: new Date(2024, 8, 14),
  parentId: 2
  subAssignments: [
    { name: 'Exercise 1',
      executionDate: new Date(2023, 9, 1),
      expiryDate: new Date(2024, 8, 14),
      subAssignments: [],
    },
}
```

jj-stigell:
> Also, could the `executionDate` be changed to `date`?

Changed to `date`

> And another question: Is it possible to add sub-attainments with sub-attainments while editing an attainment? So, that the higher attainment knows its parentId but the lower one doesn't because its parent is just being created. So that something like this would be returned form the client (only the created new sub-attainments to an existing attainment with the ID 2):
> 
> ```
> {
>   name: '2 optional exercises',
>   executionDate: new Date(2023, 9, 1),
>   expiryDate: new Date(2024, 8, 14),
>   parentId: 2
>   subAssignments: [
>     { name: 'Exercise 1',
>       executionDate: new Date(2023, 9, 1),
>       expiryDate: new Date(2024, 8, 14),
>       subAssignments: [],
>     },
> }
> ```

* Everything coming to the add attainment endpoint will be handled as a new non-existing resource.
* Everything coming to edit attainment endpoint has to exists in the database prior calling the endpoint.

New resources cannot be created while editing. If the whole resource above is new, then just send it to add attainment endpoint. "2 optional exercises" will be linked to the `parentId` with id 2 and "Exercise 1" will be automatically linked to "2 optional exercises". When creating new attainments, the "highest attainment" can be assigned `parentId` by the client (given it exists) and all sub-attainments will be automatically assigned by backend.

server/src/types/attainable.ts
```diff
@@ -0,0 +1,24 @@
+// SPDX-FileCopyrightText: 2023 The Aalto Grades Developers
+//
+// SPDX-License-Identifier: MIT
+
+export interface AttainableRequestData {
```
no-comma, line 5: This type is likely to just be used in a single file, so it could be defined in the attainable controller file, but it's not a big deal.

server/src/controllers/attainable.ts
```diff
@@ -0,0 +1,255 @@
...
+      });
+    }
+    return attainables;
+  }
+
+  const subAttainables: Array<AttainableData> = await processSubAttainables(
+    requestSubAttainables, attainable.id
+  );
+
+  res.status(HttpCode.Ok).json({
```
no-comma, line 154: I think it would be better not to have two `res.json` calls depending on whether sub attainables exist. Instead there should be a single `res.json` call at the end, with an `AttainableData` object which is modified depending on whether sub attainables exist. Pseudo-code below to illustrate what I mean.

Currently:
```ts
if (!subAttainables) {
  res.send(/* attainables */);
  return;
}

processSubAttainables(...);
res.send(/* attainables + subAttainables */);
```
Instead:
```ts
attainablesData: AttainableData = { ... };

if (subAttainables) {
  attainablesData.subAttainments = processSubAttainables(...);
}

res.send(attainablesData);
```
It may also be better to return an empty array in `attainablesData.subAttainments` if no sub attainments exist rather than leaving it undefined.

no-comma (commented):
Just two more comments. These aren't major enough to warrant blocking this PR any further if @viivi-uhari approves the PR now, but I would appreciate it if you can address them.

server/src/types/attainable.ts
jj-stigell: It is also used in test mock data, I'll keep it as is for now.

server/src/controllers/attainable.ts
jj-stigell: Good point, fixed.

no-comma (approved):
Great job!

PR #156, opened 2023-02-18
Fixes to account for backend changes

marillam:
Fixes needed to make the main functional again.

Note: I added the new ".data" to code located in the services folder for everything except the authentication related functionality. I think this is the easiest way to move forward with the least amount of changes to existing code. The auth code already worked with the new routes for the most part (outside the user service), so I decided to let it be. @kiiamakir decide if you want to change that.

And I'm now realising that the tests for fetch instances and edit instance probably fail because the mock data hasn't been changed to match the new route data.

### DIFF
--- a/client/src/components/EditInstanceView.js
+++ b/client/src/components/EditInstanceView.js
@@ -18,7 +18,7 @@ const EditInstanceView = () => {
   useEffect(() => {
     instancesService.getSisuInstance(instanceId)
       .then((data) => {
-        setInstance(data.instance);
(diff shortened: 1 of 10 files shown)


viivi-uhari (approved):
Great Job! Everything seems to work!

kiiamakir (approved):
Looks great! Currently all get and post calls seem to be working. Once the backend refactoring is completed we should make sure that the data being sent to the server is up to date.

Issue #163, opened 2023-02-22
Course role refactoring


no-comma:
As a backend developer, I want course roles to be defined better with mock data, so that old APIs can be updated to function as intended.

- [x] Teacher responsible as a user role instead of course instance attribute @no-comma
- [x] Define what course roles should exist
- [x] Update old APIs to retrieve data through course roles @no-comma
- [x] Mock data for course roles @no-comma
- [ ] Split user role and course role @Gskartwii

PR #165, opened 2023-02-24
Course role refactor

no-comma:
This PR is intended to address the tasks I've assigned to myself in #163.

Since we've received comments about it, I've opted to rename "responsible teacher" in this PR. "Teacher responsible" has been suggested, but I chose "teacher in charge" as a candidate new term. "Vastuuopettaja" doesn't strictly seem to have a translation, but "vastuuhenkilö" does, and it translates to "person in charge." By extension it would make sense for "vastuuopettaja" to be "teacher in charge."

I don't know how to handle the new queries going through `CourseRole`, so I would appreciate some pointers on those.


server/src/controllers/courseInstance.ts
```diff
@@ -71,7 +79,7 @@ export async function getCourseInstance(req: Request, res: Response): Promise<vo
     endDate: instance.endDate,
     teachingMethod: instance.teachingMethod as TeachingMethod,
     gradingType: instance.gradingType as GradingType,
-    responsibleTeacher: responsibleTeacher?.name ?? '-',
+    teachersInCharge: [],
```
jj-stigell, line 82: Can we fill in the teacher names here same way as in the `getAllCourseInstances` function?

server/src/types/course.ts
```diff
@@ -45,6 +45,13 @@ export interface CourseInstanceData {
...
-  responsibleTeachers?: Array<string>,
+  // TODO: There should maybe be a UserData interface if more data is needed,
+  // for example ID of user.
+  teachersInCharge?: Array<string>
+}
+
+export enum CourseInstanceRoleType {
+  Student = 'STUDENT',
+  Teacher = 'TEACHER',
+  TeacherInCharge = 'TEACHERINCHARGE',
```
jj-stigell, line 56: I think this is just a matter of taste , but should we use underscore as word separator in `TEACHERINCHARGE` to improve readability? (Also in the database and sequelize models)

jj-stigell (changes requested):
Nice refactoring, just few things to address.

server/src/types/course.ts
no-comma: Yes, that would be better. The reason I chose `TEACHERINCHARGE` was because it seemed more consistent with `PASSFAIL`.

jj-stigell: Would it then make sense to also update the `PASSFAIL` to `PASS_FAIL`

no-comma: Changed both. I also decided update some other naming we've decided to change before.

server/src/controllers/courseInstance.ts
no-comma: Good point, forgot about that.

jj-stigell (approved):
Good work

client/src/services/textFormat.js
```diff
@@ -22,6 +22,10 @@ const formatDateString = (dateString) => {
 // and from "EXAM" to "Exam"
 const formatCourseType = (courseType) => {
   switch (courseType) {
+  case 'teaching-participation-lectures':  // Type from Sisu API
```
Gskartwii, line 25: Should we also handle the case of `hyl/tt/ht` here? Such a grading scale is used for e.g. the Swedish language exams taken by Finnish-speaking students.

server/src/controllers/sisu.ts
```diff
@@ -23,13 +23,11 @@ function parseSisuCourseInstance(instance: SisuCourseInstance): CourseInstanceDa
...
     endDate: instance.endDate,
-    teachingMethod: (instance.type === 'exam-exam'
-      ? TeachingMethod.Exam
-      : TeachingMethod.Lecture),
-    gradingType: (instance.summary.gradingScale.fi === '0-5'
-      ? GradingType.Numerical
-      : GradingType.PassFail),
-    responsibleTeachers: instance.summary.teacherInCharge,
+    type: instance.type,
+    gradingScale: (instance.summary.gradingScale.fi === '0-5'
```
Gskartwii, line 27: Comment about language exam grading scales also applies here.

server/src/controllers/courseInstance.ts
```diff
@@ -28,14 +35,31 @@ export async function getCourseInstance(req: Request, res: Response): Promise<vo
...
+                model: CourseTranslation,
+                attributes: ['language', 'courseName', 'department']
+              }
+            ]
+          },
+          {
+            model: User,
+            attributes: ['name'],
+            where: {
+              '$Users->CourseInstanceRole.role$': CourseInstanceRoleType.TeacherInCharge
```
Gskartwii, line 57: Just checking, does Sequelize guarantee that this dollar-sign notation is valid?

server/src/controllers/courseInstance.ts
```diff
@@ -116,7 +140,7 @@ export async function getCourseInstance(req: Request, res: Response): Promise<vo
 }
 
 export interface CourseInstanceWithTeachers extends CourseInstance {
-  teacher: User
+  Users: Array<User>
```
Gskartwii, line 143: Could we consider using `{ name: string }` instead of the full `User` type here? Below, the `findAll` call only extracts the `name` attribute from the database, so this is unsound in terms of the TypeScript type system.

server/src/controllers/courseInstance.ts
```diff
@@ -125,21 +149,16 @@ export async function getAllCourseInstances(req: Request, res: Response): Promis
...
-    ],
     where: {
       courseId: courseId
     },
-    // TODO: CHANGE TO GO THROUGH COURSE ROLE INSTEAD OF GOING THROUGH
-    // RESPONSIBLE TEACHER FOREIGN KEY
     include: {
       model: User,
-      as: 'teacher'
+      attributes: ['name'],
```
Gskartwii, line 158: This would have many of the fields in the `User` being `null/undefined`, which would violate the contract of the `User` type.

Gskartwii (changes requested):
Nice work on a complex and wide-reach refactor. I have only a couple of concerns regarding a missing grading scale, and a Sequelize usage.

client/src/services/textFormat.js
viivi-uhari: Yes we can add that option. Do we have any documentation about what these types from Sisu could be? For example, is `hyl/tt/ht` confirmed to be a possible response from Sisu?

no-comma: This would fall under grading scale below rather than course instance type, but yes, it would be easy to add that to the server code, once we know the proper term for this scale.

> Do we have any documentation about what these types from Sisu could be?

One way to figure this out would be to try fetching information about a course which does use this scale from the Sisu API. Documentation probably also exists, but I don't know / can't remember where to find it.

server/src/controllers/courseInstance.ts
no-comma: Documentation for (part of?) this can be found here: https://sequelize.org/docs/v6/advanced-association-concepts/eager-loading/#complex-where-clauses-at-the-top-level

Does this answer your question? If not, could you elaborate on what you mean?

client/src/services/textFormat.js
viivi-uhari: Right, that should be added under grading scale but we have a similar function for the grading scale on the client side that we have for the course instance type so adding the correct term to the code client shouldn't be a problem either.

no-comma: The Sisu API returns this as `toinen-kotim`. Based on that, I'd add this as an enum value:
```ts
SecondLanguage = 'SECOND_LANGUAGE'
```
Any opinions?

viivi-uhari: That seems to be a good translation, or `'SECOND_NATIVE_LANGUAGE'`. But since this is used to return the values 'hyvät taidot', 'tyydyttävät taidot' etc., I'm not sure if the term used in Sisu is the best. But if we want to follow the example from Sisu, I'd say `'SECOND_LANGUAGE'` or `'SECOND_NATIVE_LANGUAGE'` would work.

viivi-uhari: Here are some terms that came to mind, but I feel like using some kind of `LANGUAGE` type would be fine as well.

Level of understanding
Level of skills
skill level
Level of abilities
Ability
Competence

viivi-uhari:
Because of the grading scale conversation I actually started thinking that have all of the Sisu course types been accounted for on the server side. For example, in the Sisu UI the type for the courses CS-C3170 and CS-E4675 is 'Project'.

viivi-uhari:
> Based on the terminology meeting, we aren't considering any Sisu course or course instance types in the backend at all anymore in this PR.

Right, I forgot about this! Then everything is good!

server/src/controllers/courseInstance.ts
Gskartwii: That explains it! I was wondering if this was an undocumented hack at first, but it's all good :+1:

client/src/services/textFormat.js
no-comma: Something needs to be picked so I chose `SECOND_NATIONAL_LANGUAGE`, which can be changed later.

viivi-uhari: I then need to add this to frontend as well. So hold on the review @Gskartwii

Gskartwii (approved):
Good work!

Issue #134, opened 2023-02-01
Grading formula selection with sample formula


no-comma:
As a teacher I want to select a formula from menu to calculate the final grade from assignments on my course. If previous instance exists, the formula (and assignments) can be copied from there.

To be able to test this more concretely, we could test with the arrangement that was used on C course: Weighed average of "subgrades" (floating point value between 0.5 and 5.5), that is rounded to nearest integer. Additionally each assignment category has a threshold requirement (e.g., must reach at least 50% of available points) that is required for acceptable grade.

Probably the best place to choose the formula is on the course instance view where assignment names and parameters are edited, because the formula determines what parameters there are. For example, in this case each assignment type has three attributes that can be modified: max points, min required points, and weight (between 0 and 1, or 0 - 100%).

DoD: the above case is demonstrated and tested. Adding new formulas should be relatively easy (e.g. a single function with well-specified generic interface), but formula editor is not needed at this point.

- [x] Frontend design for formula selection 
- [x] Frontend implementation for formula selection
- [ ] API for getting existing formulas

PR #166, opened 2023-02-26
Weighted average calculation

Gskartwii:
I've been sketching up an API for grade calculation iteratively. This is all in very early stages, and feedback is welcome.


Gskartwii:
Related: #134

server/src/types/assignment.ts
```diff
@@ -0,0 +1,52 @@
...
+}
+export type ManuallyGradedParams = {
+  min: number;
+  max: number;
+}
+
+export abstract class Assignment<P extends AssignmentParams> {
+  public abstract calculateGrade(
+    subGrades:
+      this extends ManuallyGradedAssignment ? number: Array<number>
```
Gskartwii, line 17: Instead of `Array<number>`, how about:
```typescript
class Attainment<P extends AssignmentParams> {
    public assignment: Assignment<P>;
    public subAttainments: Array<Attainment<???> | number>;
}
```
But I'm unsure about what to put in place of `???` for the generic params.

Gskartwii:
Ready for review. The testing could be more thorough, so feel free to ask for more tests if you feel it is lacking.

no-comma (changes requested):
A general comment I have is that since a lot of the code in this PR is quite complex, it could be commented very liberally.

server/src/database/index.ts
```diff
@@ -25,7 +25,7 @@ export const sequelize: Sequelize = new Sequelize(
       underscored: true,
       freezeTableName: true
     },
-    logging: console.log
+    logging: false//console.log
```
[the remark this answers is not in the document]
Gskartwii: You're right! I will revert this change.

server/src/routes/attainable.ts
```diff
@@ -53,6 +53,24 @@ export const router: Router = Router();
...
  *       - $ref: '#/definitions/AddAndEditAttainment'
+ *   Grades:
+ *     description: Calculated final grades for each student.
+ *     type: array
+ *     items:
+ *       type: object
+ *       properties:
+ *         studentId:
+ *           type: number
+ *           description: Student database identifier.
```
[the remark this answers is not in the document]
Gskartwii: Good point, I think we should hold off on finalizing this PR until the database representation is clear.

server/src/types/assignment.ts
```diff
@@ -0,0 +1,52 @@
+// SPDX-FileCopyrightText: 2023 The Aalto Grades Developers
+//
+// SPDX-License-Identifier: MIT
+
+export interface AssignmentParams {
+  min: number;
+  max: number;
+}
+export type ManuallyGradedParams = {
```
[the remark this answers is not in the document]
Gskartwii: This file was in fact unused, good catch, I decided to just remove this file as it is not necessary in our revised design.

server/src/types/assignment.ts
```diff
@@ -0,0 +1,52 @@
...
+  min: number;
+  max: number;
+}
+
+export abstract class Assignment<P extends AssignmentParams> {
```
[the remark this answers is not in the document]
Gskartwii: Yes, I'll delete this file.

server/src/controllers/attainable.ts
```diff
@@ -233,3 +236,191 @@ export async function updateAttainable(req: Request, res: Response): Promise<voi
...
+  weights: Array<number>;
+}
+
+type FormulaFunction = (subResults: Array<CalculationResult>) => Promise<CalculationResult>;
+type ParameterizedFormulaFunction = (parameters: any, subResults: Array<CalculationResult>) => Promise<CalculationResult>;
+type FormulaNode = {
+  validatedFormula: FormulaFunction;
+  subFormulaNodes: Array<FormulaNode>;
+};
+const formulasWithSchema: Map<Formula, [yup.AnyObjectSchema, ParameterizedFormulaFunction]> = new Map();
```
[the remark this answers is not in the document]
Gskartwii: How would you feel about `/types`? Or would this perhaps warrant a `/services` entry?

server/src/controllers/attainable.ts
```diff
@@ -233,3 +236,191 @@ export async function updateAttainable(req: Request, res: Response): Promise<voi
...
+  ]
+);
+
+async function calculatedWeightedAverage(
+  params: WeightedAssignmentParams,
+  subResults: Array<CalculationResult>
+): Promise<CalculationResult> {
+  const weighted: CalculationResult =
+    params.weights
+      .reduce(
```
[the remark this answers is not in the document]
Gskartwii: I can see the benefit of a loop here, especially when we can benefit from early exit.

server/src/controllers/attainable.ts
```diff
@@ -233,3 +236,191 @@ export async function updateAttainable(req: Request, res: Response): Promise<voi
...
+        (
+          acc: CalculationResult,
```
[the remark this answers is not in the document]
Gskartwii: `acc` stands for "accumulator", it is quite typically used in the context of `reduce`.

server/src/routes/attainable.ts
```diff
@@ -160,3 +178,38 @@ router.put(
...
+ *             schema:
+ *               $ref: '#/definitions/Failure'
+ *       404:
+ *         description: The given course or course instance does not exist.
+ *         content:
+ *           application/json:
+ *             schema:
+ *               $ref: '#/definitions/Failure'
+ */
+router.get(
```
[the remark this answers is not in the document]
no-comma: Since this route is likely to save data to the database in the future, we discussed that `POST` would probably be the best method.

server/src/routes/attainable.ts
```diff
@@ -160,3 +178,38 @@ router.put(
   handleInvalidRequestJson,
   controllerDispatcher(updateAttainable)
 );
+
+/**
+ * @swagger
+ * /v1/courses/{courseId}/instances/{instanceId}/grades:
```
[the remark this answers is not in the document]
no-comma: We thought that the route for this functionality could be: `POST /v1/courses/{courseId}/instances/{instanceId}/grades/calculate`

As such I think that the route and controller of this PR could be moved to new `grades.ts` files.

server/src/controllers/attainable.ts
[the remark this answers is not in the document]
no-comma: The definition of formula related types is fine under `src/types`, but the formula functions themselves aren't exactly types. We discussed that we could have a new directory `src/formulas` where formulas are stored, since `services` sounds ambiguous. In the future there may also be a lot of formulas, so formulas having their own directory would be justified.

server/src/formulas/index.ts
```diff
@@ -0,0 +1,21 @@
...
+registerFormula(
+  Formula.Manual,
+  yup.object(),
+  // If no points have been input for a student, assume the attainment
+  // has been failed.
+  async (
+    _params: any,
+    _subGrades: Array<CalculationResult>,
+  ): Promise<CalculationResult> => {
+    return { status: Status.Fail, points: undefined };
```
no-comma, line 18: Again I personally prefer
```ts
return {
  status: Status.Fail,
  points: undefined
}
```

server/src/controllers/grades.ts
```diff
@@ -0,0 +1,158 @@
...
+  formulaChecker,
+  FormulaNode,
+  FormulaParams,
+  getFormula,
+  Status
+} from '../types/formulas';
+import { idSchema } from '../types/general';
+import { HttpCode } from '../types/httpCode';
+
+async function calculate(
```
no-comma, line 23: More descriptive name. Calculate what and/or how?

server/src/controllers/grades.ts
```diff
@@ -0,0 +1,158 @@
...
+  tree: FormulaNode,
+  presetPoints: Map<FormulaNode, number>
+): Promise<CalculationResult> {
+  if (presetPoints.has(tree)) {
```
no-comma, line 27: Comment to explain this is to handle cases where a teacher has manually specified a grade for an attainment.

server/src/controllers/grades.ts
```diff
@@ -0,0 +1,158 @@
...
+    return { status: Status.Pass, points: presetPoints.get(tree) };
+  }
+  // Not calculated so far, so calculate the points for lower nodes.
```
no-comma, line 30: "Not calculated" might be misleading, more commonly this would happen when a grade hasn't been manually specified for this attainment.

server/src/controllers/grades.ts
```diff
@@ -0,0 +1,158 @@
...
+  const subPoints: Array<CalculationResult> =
+    await Promise.all(
+      tree.subFormulaNodes.map(
+        (subTree: FormulaNode) => calculate(subTree, presetPoints)
+      )
+    );
+  // Based on the results from lower nodes, calculate the points for this node.
```
no-comma, line 37: ...using the formula specified for this attainment/node.

server/src/controllers/grades.ts
```diff
@@ -0,0 +1,158 @@
...
+export async function calculateGrades(
+  req: Request,
+  res: Response
+): Promise<void> {
+  const courseId: number = Number(req.params.courseId);
+  const courseInstanceId: number = Number(req.params.instanceId);
+  await idSchema.validate({ id: courseId }, { abortEarly: false });
+  await idSchema.validate({ id: courseInstanceId }, { abortEarly: false });
+
+  const formulaNodesById: Map<number, FormulaNode> = new Map();
```
no-comma, line 50: What ID?

server/src/controllers/grades.ts
```diff
@@ -0,0 +1,158 @@
...
+
+  const attainables: Array<{
```
no-comma, line 52: This should just be changed to `attainments`. See #138 and discussion on Teams.

server/src/controllers/grades.ts
```diff
@@ -0,0 +1,158 @@
...
+    id: number,
+    attainableId: number,
```
no-comma, line 54: IIRC this is a Sequelize issue forcing the use of the name "attainableId", but here it would be good if you can somehow specify that this is specifically the parent attainment's ID.

server/src/controllers/grades.ts
```diff
@@ -0,0 +1,158 @@
...
+      'formulaParams',
+    ]
+  });
+
+  let rootAttainable: null | FormulaNode = null;
+  for (const attainable of attainables) {
+    const formulaId: Formula | null = attainable.formulaId;
+    const params: FormulaParams | null = attainable.formulaParams;
+    if (!(await formulaChecker.validate(formulaId))) {
+      throw new Error('bad');
```
no-comma, line 75: Needs a better error message.

server/src/controllers/grades.ts
```diff
@@ -0,0 +1,158 @@
...
+      throw new ApiError('the parameters for a formula haven\'t been set', HttpCode.BadRequest);
+    }
+
+    formulaNodesById.set(attainable.id, {
+      validatedFormula: await getFormula(formulaId as Formula, params),
+      subFormulaNodes: [],
+    });
+  }
+
+  for (const attainable of attainables) {
```
no-comma, line 87: Explain what this loop does.

no-comma (changes requested):

server/src/controllers/grades.ts
```diff
@@ -88,3 +95,324 @@ export async function addGrades(req: Request, res: Response, next: NextFunction)
...
+   * the manually specified grade will be used.
+   *
+   * A grade has to be manually specified when using the 'Manual' formula
+   * and a grade may be manually specified for overriding grade calculation
+   * functions in special cases.
+   */
+  const presetGrade: number | undefined = presetGrades.get(formulaNode);
+  if (presetGrade) {
+    return {
+      status: Status.Pass,
```
Gskartwii, line 120: Can the teacher tell the system that a student has in fact failed an assignment?

server/src/controllers/grades.ts
```diff
@@ -88,3 +95,324 @@ export async function addGrades(req: Request, res: Response, next: NextFunction)
...
+    },
+    attributes: [
+      'id',
+      'attainableId',
+      'formula',
+      'parentFormulaParams',
+    ],
+  })).map(
+    // TODO: Remove this map() call.
+    // map() is called as a temporary measure to translate attainableId to parentId.
```
Gskartwii, line 183: We might be able to perform this translation using [Sequelize API](https://sequelize.org/api/v6/class/src/model.js~model#static-method-findAll). Perhaps in the attribute list above, replace `'attainableId'` with `['attainableId', 'parentId']`, if I have understood the docs correctly.

server/src/controllers/grades.ts
```diff
@@ -88,3 +95,324 @@ export async function addGrades(req: Request, res: Response, next: NextFunction)
...
+    (attainment: Attainable) => {
```
Gskartwii, line 184: If we decide to keep this map call or a similar construct, I would like to avoid having `Attainable` as the type here. Certain properties of `Attainable` are undefined, going against the interface of `Attainable`.

Gskartwii (commented):
Excellent documentation, the code is easy to follow.

Gskartwii:
I'm going to write some unit tests for the formulas right now.

server/src/controllers/grades.ts
no-comma: True, that should be possible in the future, but I think we can leave it out of this PR.

server/src/controllers/grades.ts
no-comma: Can you try it? If it works, we can do this that way. But in either case, this would hopefully only be temporary.

Gskartwii: Yes, it seems to work. There were a few other [tweaks](https://github.com/aalto-grades/base-repository/pull/166/commits/c4dd3ad37badbbecbc95898fc78737db5d7dce87) to enable this mapping, though.

server/test/formulas/weightedAverage.test.ts
```diff
@@ -0,0 +1,63 @@
...
+      { params: { min: 0, max: 20, weight: 0.3 }, subResult: { grade: 10, status: Status.Pass } },
+      { params: { min: 0, max: 20, weight: 0.7 }, subResult: { grade: 14, status: Status.Pass } },
+      { params: { min: 0, max: 3, weight: 1 }, subResult: { grade: 3, status: Status.Pass } },
+    ];
+    const computedGrade: GradingResult = await implementation.formulaFunction(input);
+    expect(computedGrade.grade).toBeCloseTo(15.8);
+    expect(computedGrade.status).toBe(Status.Pass);
+  });
+
+  it('should calculate a failing grade when a subgrade is failing', async () => {
```
no-comma, line 51: We may not want to assume an attainment is failed because one of its subattainments is failed, but it's fine for now.

server/src/controllers/grades.ts
```diff
@@ -354,3 +357,322 @@ export async function addGrades(req: Request, res: Response, next: NextFunction)
...
+    if (attainment.parentId === null) {
+      /*
+       * This attainment is the root attainment. Store a reference to its formula
+       * node in the rootFormulaNode variable.
+       */
+
+      // There should only be one root attainment.
+      if (rootFormulaNode) {
+        throw new ApiError(
+          'duplicate root attainment',
```
jj-stigell, line 534: What if there are two separate attainment "trees" on the instance, one for example with exam attainments and other one with A+ attainments? Then there will be 2 attainments that don't have parents but only children, and this error is raised.

jj-stigell (changes requested):
Looks good for the first iteration. Few questions that might need some actions/clarifications.

server/src/types/formulas.ts
```diff
@@ -0,0 +1,48 @@
...
+import * as yup from 'yup';
+
+export enum Formula {
+  Manual = 'MANUAL',
+  WeightedAverage = 'WEIGHTED_AVERAGE',
+}
+
+export enum Status {
+  Pass = 'pass',
+  Fail = 'fail',
```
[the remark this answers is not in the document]
no-comma: Probably, but I'm not completely sure if this enum is even necessary. It may be sufficient to just have `passed: boolean` in `GradingResult`.

@Gskartwii considered the possibility of adding other statuses in the future such as "grading in progress," but I don't know if that's relevant in this context. Maybe if it is relevant in another context, this may still be good to have in grade calculation to be consistent with other uses.

server/src/controllers/grades.ts
```diff
@@ -354,3 +357,322 @@ export async function addGrades(req: Request, res: Response, next: NextFunction)
...
+
+  /*
+   * First find the formula implementation and formula parameters for the parent
+   * attainment's formula.
+   */
+  for (const attainment of attainments) {
+    const formula: Formula | null = attainment.formula;
+
+    // Ensure that the formula specified for this attainment is valid.
+    if (!yup.string().oneOf(Object.values(Formula)).required().validate(formula)) {
```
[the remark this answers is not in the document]
no-comma: I can't remember the exact reason for including this check, but it may very well be redundant.

server/src/controllers/grades.ts
```diff
@@ -354,3 +357,322 @@ export async function addGrades(req: Request, res: Response, next: NextFunction)
...
+   * Attainment ID -> Formula node of the given attainment.
+   */
+  const formulaNodesByAttainmentId: Map<number, FormulaNode> = new Map();
+
+  /*
+   * First find the formula implementation and formula parameters for the parent
+   * attainment's formula.
+   */
+  for (const attainment of attainments) {
+    const formula: Formula | null = attainment.formula;
```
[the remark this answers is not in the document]
no-comma: Yes probably, just remove the option for a null formula from `AttainmentInfo`.

server/src/controllers/grades.ts
jj-stigell: Ok, so the exam and other possible attainment trees are stored under the one root attainment. 

Good point on the final exam vs. course exam and other points. If I remember correctly same thing was briefly discussed on one meeting last week with teaching staff.

But as this is the first iteration of grade calculation, I think we can ignore this for now and update later if needed.

no-comma: Actually even with final exams, we wouldn't need two root attainments. We could still have a single root attainment for the final grade with a formula such as "choose best," and under that, there are two attainments, one for the final exam and the other for regular course completion.

But this should be discussed in more detail later.

PR #167, opened 2023-03-01
Edit assignment view

viivi-uhari:
Frontend hasn't incorporated the new terminology yet so I'll be using the phrase 'assignment'.

Initial edit assignment view page created. Not connected to backend yet. All buttons directing to this page (and out) should work except this still need to be connected with the final formula selection pages. Also, `mockAssignmentsClient.js` has been improved regarding grading formulas. 

Now when assignments are created, the properties `affectCalculation` and `formulaAttributes` are added to each assignment, but `formulaId` is not. These properties can be removed/added later if necessary when connecting to backend. These Properties are initialised with empty values so creating/omitting them now doesn't make a difference since this page isn't connected to backend yet.

Performance needs to be improved: when too many text fields are displayed at once, typing is slow.


client/src/components/AddAssignmentsView.js
```diff
@@ -75,17 +75,17 @@ const AddAssignmentsView = () => {
                 <AssignmentCategory 
                   key={assignment.id} 
                   assignment={assignment}  
-                  button={<Button>Edit</Button>} 
+                  button={<Button onClick={ () => navigate('/edit-assignment/' + instanceId + '/' + assignment.id) }>Edit</Button>}
```
kiiamakir, line 78: Could some of this be separated onto individual lines to shorten the line-length

client/src/components/AddAssignmentsView.js
```diff
@@ -75,17 +75,17 @@ const AddAssignmentsView = () => {
...
                 />
               );}
             ) }
           </Box>
         }
-        <Typography variant='body2' color='primary.main' sx={{ m: '8px 0px' }} >You can also add assignments after creating the instance</Typography>
+        <Typography variant='body2' color='primary.main' sx={{ m: '8px 0px' }} >You can also add study attainments after creating the instance</Typography>
```
kiiamakir, line 84: Same with here, typography text onto its own line

client/src/components/CreateAssignmentView.js
```diff
@@ -11,13 +12,17 @@ import Assignment from './create-assignment/Assignment';
...
 const CreateAssignmentView = () => {
+  const navigate = useNavigate();
 
   // The property 'category' must be specified for each assignment in order to populate the textfields correctly
   const [assignments, setAssignments] = useState([{
     category: '',
     name: '',
     date: '',
     expiryDate: '',
+    //formulaId: null,
```
kiiamakir, line 23: Delete this if not needed?

client/src/components/create-assignment/ConfirmationDialog.js
```diff
@@ -16,14 +16,14 @@ import Button from '@mui/material/Button';
 function ConfirmationDialog({ handleClose, open, removeAssignment, indices, assignments }) {
   return (
     <Dialog open={open} >
-      <DialogTitle >Delete sub-assignment</DialogTitle>
+      <DialogTitle >Delete Sub-Attainments</DialogTitle>
```
kiiamakir, line 19: Are we calling these sub-attainments or sub study attainments? There are some inconsistencies with this in Figma, so we should decide on one wording

client/src/components/create-assignment/ParentAssignment.js
```diff
@@ -27,6 +30,17 @@ const ParentAssignment = ({ indices, addSubAssignments, setAssignments, assignme
...
 
+  /* Functions to get the formula attributes.
+     formulaId specifies the formula that is used to calculate this assignmnet's garde,
+     subFormulaAttributeNames are the attributes that need to be specified for the direct sub assignments of this assignments,
+     so that the grade for this assignment can be calculated. 
+     Observe that formulaAttributeNames that is as a parameter for this component are the attributes that need to specified for this assignmnet,
+     so that the grade of this assignment's parent assignment can be calculated.
+  */
+  const formulaId = assignmentServices.getProperty(indices, assignments, 'formulaId');
+  const formulaName = formulasService.getFormula(formulaId);
```
kiiamakir, line 41: Should this getFormula return the whole formula as an object and not just the name?

client/src/components/create-assignment/ParentAssignment.js
```diff
@@ -73,6 +88,7 @@ const ParentAssignment = ({ indices, addSubAssignments, setAssignments, assignme
                   assignments={assignments} 
                   setAssignments={setAssignments} 
                   removeAssignment={removeAssignment}
+                  formulaAttributeNames={subFormulaAttributeNames ? subFormulaAttributeNames : []}
```
kiiamakir, line 91: I believe that subFormulaAttributeNames ?? [] should work here

kiiamakir (changes requested):
Good stuff! We should consider adding a line length check to eslint (backend already has this) to make it easier to spot lines that are too long.

client/src/components/create-assignment/ParentAssignment.js
viivi-uhari: Thanks for the tip but if I use this then I get the following error: `Invalid prop `formulaAttributeNames` of type `string` supplied to `Assignment`, expected `array`.`

client/src/components/create-assignment/ParentAssignment.js
viivi-uhari: I could actually change the function name to `getFormulaName`?

client/src/services/textFormat.js
```diff
@@ -9,6 +9,12 @@ const formatDateToString = (date) => {
   return string.split(',')[0];
 };
 
+// Format Date type values to strings of the format "2023-01-01"
```
marillam, line 12: Purely a comment: Should we look into our dates more closely? I feel like we have quite a few date formatting functions and I think it might be confusing one day when new people start working on this.

client/src/components/create-assignment/ConfirmationDialog.js
viivi-uhari: Decided with the frontend team to use 'sub study attainment' when there's room, so almost always, but in buttons and input labels we use 'sub attainment'.

client/src/services/textFormat.js
viivi-uhari: Yeah we could talk about this. There are different formats coming in from the server (from our instances and Sisu instances) and also different formats are needed for the UI, so that's why there are many functions to format them.

marillam (approved):

kiiamakir (approved):

PR #168, opened 2023-03-02
Include course ID in all course instance routes

no-comma:
Adds a requirement to include course ID in the remaining course instance route which doesn't require it.

Ideally we would use composite keys, but there seemed to be problems with getting Sequelize to work with them cleanly. But that can be addressed later and hopefully without breaking the frontend connection after it has been fixed in this PR.

### DIFF
--- a/server/src/controllers/courseInstance.ts
+++ b/server/src/controllers/courseInstance.ts
@@ -47,8 +47,9 @@ export async function getCourseInstance(req: Request, res: Response): Promise<vo
   // Verify that the course instance belongs to the found course.
   if (instance.courseId != courseId) {
     throw new ApiError(
-      `course instance with ID ${instanceId} belonging to course with ID ${courseId} not found`,
-      HttpCode.NotFound
+      `course instance with ID ${instanceId} ` +
+      `does not belong to the course with ID ${courseId}`,
+      HttpCode.Conflict
     );
   }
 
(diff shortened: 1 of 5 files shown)


server/src/controllers/courseInstance.ts
```diff
@@ -17,45 +17,41 @@ import { HttpCode } from '../types/httpCode';
...
     throw new ApiError(
       `course instance with ID ${instanceId} not found`, HttpCode.NotFound
     );
   }
 
+  // Verify that the course instance belongs to the found course.
+  if (instance.courseId != courseId) {
+    throw new ApiError(
+      `course instance with ID ${instanceId} belonging to course with ID ${courseId} not found`,
+      HttpCode.NotFound
```
jj-stigell, line 51: In PR #153 we use following message with http conflict when instance does not belong to the course.
```javascript
// Check that instance belongs to the course.
if (instance.courseId !== course.id) {
  throw new ApiError(
    `course instance with ID ${courseInstanceId} ` +
    `does not belong to the course with ID ${courseId}`,
    HttpCode.Conflict
  );
}
```

Should we use this here or update other PR to match this code?

jj-stigell (changes requested):

server/src/controllers/courseInstance.ts
no-comma: Tough to say. Ideally we wouldn't need to do this check at all, but I can change this to 409 for now.

jj-stigell (approved):
👍

jyrge (approved):

PR #170, opened 2023-03-08
Initial robot framework tests

lakooi:


no-comma:
The tests now run in GitHub Actions, although all of the actual test cases are skipped. But I think those can be fixed in a later PR.

server/Dockerfile
```diff
@@ -6,11 +6,11 @@ FROM node:16-alpine
 
 WORKDIR /app
 
-ARG SERVER_SKIP_CI
+ARG SERVER_SKIP_NPM_CI
 
 COPY --chown=node:node . .
 
-RUN if [ -z "$SERVER_SKIP_CI" ] || [ $SERVER_SKIP_CI != 1 ]; then npm ci; fi
+RUN if [ -z "$SERVER_SKIP_NPM_CI" ] || [ $SERVER_SKIP_NPM_CI != 1 ]; then npm ci; fi
```
jj-stigell, line 13: instead of using the `SERVER_RUN_TESTS` and `SERVER_SKIP_NPM_CI` to check if we are running tests , could we just utilize the `NODE_ENV` here, like `NODE_ENV: "test"`. Other backend functionality uses this to determine the running environment.

Something like:
* `CMD ["/bin/sh", "-c", "npm run migration:up && npm run seed:up && if [ $NODE_ENV == "test"]; then npm test; else npm run start; fi"]
`
* `RUN if [ -z "$NODE_ENV" ] || [ $NODE_ENV != "test" ]; then npm ci; fi`

jj-stigell (approved):
LGTM

Just one question about the environment variables, if this change is not possible or I understood something incorrectly then you can dismiss the question.

server/Dockerfile
no-comma: For `SERVER_RUN_TESTS` that should work, but `SERVER_SKIP_NPM_CI` is purely build time functionality unrelated to whether tests are being run. It's a convenience for the developer so they don't need to wait for the dependencies to install every time they build the backend container.

jj-stigell: Thanks for the update and clarification

common-services.yaml
```diff
@@ -0,0 +1,50 @@
...
+# Docker Compose files
+
+version: "3.7"
+
+services:
+  frontend:
+    build: ./client
+    restart: always
+    ports:
+      - 3005:3005
```
lakooi, line 15: This is actually something that could be left out of the template, as we do not need exposition between localhost and the container in use cases outside of local development

lakooi: Similarly to the other services

robot-tests/authentication/signUpAndSignIn.robot
```diff
@@ -0,0 +1,24 @@
...
+Test SignUp on localhost
+    Skip
+    Open Aalto Grades on Localhost
+    Navigate to SignUp
+    Fill Admin Register Information
+    Verify Admin User registering succeeded
+    Sign Out
+
+Test Login on localhost
+    Skip
```
lakooi, line 20: There is no need to skip these tests, as they are functional

lakooi: Test Login on localhost and Test SignUp on localhost

no-comma: My mistake, I thought they failed in GitHub Actions but I had skipped the sign up test so the login failed by default.

jyrge (approved):

PR #171, opened 2023-03-09
Formula attributes view

kiiamakir:
Formula attributes view without connection to backend. Formula selection also slightly modified and connected to course view.


client/src/components/select-formula-view/SelectFormulaForm.js
```diff
@@ -114,12 +167,12 @@ const SelectFormulaForm = ({ assignments, formulas, courseId }) => {
           flexWrap: 'wrap',
           justifyContent: 'space-between'
         }}>
-          <Typography width={350} m={3}>Specify attribute values for the sub study attainnments</Typography>
+          <Typography width={350} m={3}>Specify attribute values for the sub study attainments</Typography>
```
viivi-uhari, line 170: With my screen max width, if the with of the typography is 350, the buttons will displayed under it. I tested that with the width 320, the buttons will be displayed next to it and the format of the typography doesn't change, so maybe that could be changed.

client/src/components/select-formula-view/SelectFormulaForm.js
```diff
@@ -114,12 +167,12 @@ const SelectFormulaForm = ({ assignments, formulas, courseId }) => {
...
           <Box sx={{ m: 3, mt: 0, alignSelf: 'flex-end',display: 'flex', lexDirection: 'column', }}>
-            <Button size='small' variant='outlined' type='submit' name='skipAttributes' sx={{ mr: 2 }}>
+            <Button size='medium' variant='outlined' type='submit' name='skipAttributes' sx={{ mr: 2 }}>
```
viivi-uhari, line 172: Could a 'Cancel' option also be added (with `onClick => navigate(-1)`)? Because now I kind of expect that with the 'Skip for Now' button I could also cancel selecting a formula but I actually can't since I need to specify the formula and at least one sub attainment to be able to press that button.

client/src/components/formula-attributes-view/Assignment.js
```diff
@@ -0,0 +1,70 @@
...
+
+  return (
+    <Box sx={{
+      bgcolor: '#FFFFFF',
+      display: 'flex',
+      flexDirection: 'column',
+      boxShadow: 2, 
+      borderRadius: 2,
+      px: 3,
+      py: 1,
```
viivi-uhari, line 41: I would have liked a bit more padding on the bottom so I would change this to: `pt: 1, pb: 1.5`. But I guess this depends on what you like.

client/src/components/select-formula-view/StyledBox.js
```diff
@@ -0,0 +1,14 @@
...
+//
+// SPDX-License-Identifier: MIT
+
+import Box from '@mui/material/Box';
+import styled from 'styled-components';
+
+const StyledBox = styled(Box)`
+width: 53vw;
+min-width:  400px;
+max-width: 1000px;
```
viivi-uhari, line 11: Indentions of 2 spaces could be added here even though the linter doesn't spot them.

viivi-uhari (changes requested):
Look great! A couple of small changes could be made and a 'cancel' button added.

client/src/components/course-view/Assignments.js
```diff
@@ -11,15 +11,15 @@ import Typography from '@mui/material/Typography';
...
+const Assignments = ({ assignments, formula, instance, courseCode }) => {
   const navigate = useNavigate();
   
   return (
     <Box borderRadius={1} sx={{ bgcolor: 'primary.light', p: 1.5, display: 'inline-block' }}>
       <Typography variant='h6' align='left' sx={{ ml: 1.5 }} >Study Attainments</Typography>
       <Box sx={{ display: 'flex', flexWrap: 'wrap', justifyContent: 'space-between', alignItems: 'center', pb: 1 }}>
         <Typography align='left' sx={{ ml: 1.5 }} >{'Grading Formula: ' + formula}</Typography>
-        <Button onClick={ () => navigate('/select-formula') }>Edit formula</Button>
+        <Button onClick={() => navigate(`/${courseCode}/select-formula/${instance.id}`) }>Edit formula</Button>
```
viivi-uhari, line 22: A comment could be added here that one the `courseId` can be fetched from the path instead of `courseCode`, then this should be changed.

viivi-uhari (changes requested):

client/src/components/select-formula-view/SelectFormulaForm.js
kiiamakir: Cancel button added to the bottom of the form.

client/src/components/select-formula-view/SelectFormulaForm.js
```diff
@@ -16,64 +16,118 @@ import InputLabel from '@mui/material/InputLabel';
...
   const handleFormulaChange = (event) => {
     const newFormula = formulas.find(formula => formula.name == event.target.value);
     setCodeSnippet(newFormula.codeSnippet);
-    setFormula(event.target.value);
+    setSelectedFormula(newFormula);
   };
 
-  const handleSubmit = (event) => {
+  // checks that user has selected a function and at least one attainment
+  // if not, shows error message
```
marillam, line 71: Should this error be handled as a message in the form? Alert could be used if the submitting goes wrong for some reason when the form is okay. Not a mandatory fix but just an idea.

marillam (changes requested):

client/src/components/select-formula-view/SelectFormulaForm.js
kiiamakir: The error messages have been changed to be showed in the form

marillam (approved):

viivi-uhari (approved):

PR #172, opened 2023-03-10
Fix coverage reporting in server Dockerfile

Gskartwii:
This re-enables coverage reporting, which has been plagued by permission problems for a while

### DIFF
--- a/server/Dockerfile
+++ b/server/Dockerfile
@@ -14,7 +14,7 @@ RUN if [ -z "$SERVER_SKIP_CI" ] || [ $SERVER_SKIP_CI != 1 ]; then npm ci; fi
 
 RUN npm run build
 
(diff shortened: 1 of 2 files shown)


server/Dockerfile
```diff
@@ -14,6 +14,8 @@ RUN if [ -z "$SERVER_SKIP_CI" ] || [ $SERVER_SKIP_CI != 1 ]; then npm ci; fi
 
 RUN npm run build
 
+RUN mkdir /app/coverage && chown -R node:node /app/coverage
```
no-comma, line 17: Can we run `mkdir` before `COPY` so `chown` doesn't need to be repeated?

no-comma (approved):

jj-stigell (approved):

PR #173, opened 2023-03-11
Backend Attainments Connection

viivi-uhari:
Connect edit and create attainment view to backend, connect edit and create attainment view to local context, fix path names and refactor variables’ and functions’ names.
When edit and create attainment views are accessed via the course page, these views are connected to backend and the data in the database will be updated. The paths used for these views are then 
```
<Route path=':courseId/create-attainment/:instanceId' element={<CreateAssignmentView/>}/>
<Route path=':courseId/edit-attainment/:instanceId/:attainmentId' element={<EditAssignmentView/>}/>
```
The creation and editing of attainment works if you use the right paths and mock data. However, the views that are located behind buttons that bring to these views, don’t send data of the correct format to the server (for example, because the courseId is still a string and not a number, and routes for getting attainments aren’t completed yet).

When edit and create attainment views are accessed during the instance creation, these views aren’t connected to backend and they update the local context instead. The attainments need to be created when the actual instance is created. The paths used for these views are
```
<Route path=':courseId/create-temporary-attainment/:sisuInstanceId' element={<CreateAssignmentView/>}/>
<Route path=':courseId/edit-temporary-attainment/:sisuInstanceId/:attainmentId' element={<EditAssignmentView/>}/>
```

When getting the attainments for the add attainments view during instance creation, I give the attainments temporary Ids so that they can be edited, deleted, and distinguished from each other in accordions. This is done by creating a new property for the attainments: `temporaryId`. This id is fetched from the local context and incremented every time a new attainment is created so that same ids aren’t used. 

We don’t care about the Ids used on the server because we are only getting the old attainments to use as templates for the suggested attainments. These suggested attainments can then be edited and so on so the old attainments aren’t used directly under the new instance but instead new attainments are created base on them.

A new `temporaryId` property is created instead of rewriting the (already possibly existing) `id` property, because when editing an attainment, the id property tells which attainments already exist in the database and which don’t (the attainments that have ids, exist; the attainments that don’t have ids, don’t exist). And because the `EditAssignmentView` is used when editing attainments that partly exist in the DB, the difference between `id` and `temporaryId` is important. Because of these two types of ids, the attainments accordions use the `temporaryId` property as keys during instance creation, and the actual attainment `id`s as keys in the course view.

Because these temporary Ids have now been partly created, at some point we could possibly switch from using indices in the create and edit attainment views to using this `temporaryId` property instead, but I didn’t make that change yet since it could take some time. 
Finally, at the moment, suggested attainments are fetched every time again when `addedAttainments` for the new instance is empty, but this could possibly changed. However, I couldn’t figure out an easy way to stop this unnecessary fetching, so I think this again could be addressed later on.


marillam (approved):
What a branch! This looked like a lot of work, great job for being so thorough with the changes. I did think during the review that -in the future- we should separate any big term changes to their own pr. It was difficult to focus on actual changes to the logic since so many changes came from refactoring. 

In general I think we should merge this branch like this (unless @kiiamakir finds some important errors). Merging this with my pull requests will require likely require effort to ensure everything goes well, but I think it's best to tackle that is an actual merge instead of trying to fit everything beforehand. Most of the paths seemed to match already though, so we should be good.

client/src/components/create-assignment/LeafAssignment.js
```diff
@@ -15,7 +15,7 @@ import StringTextField from './StringTextField';
 import DateTextField from './DateTextField';
 import assignmentServices from '../../services/assignments';
 
-// An Assignmnet component without subassignments and hence without a formula as well.
+// An Assignmnet component without subAttainments and hence without a formula as well.
```
kiiamakir, line 18: Still assignment written here

client/src/components/create-assignment/ParentAssignment.js
```diff
@@ -18,12 +18,12 @@ import Assignment from './Assignment';
 import assignmentServices from '../../services/assignments';
 import formulasService from '../../services/formulas';
 
-// An Assignmnet component with subassignments and a formula
+// An Assignmnet component with subAttainments and a formula
```
kiiamakir, line 21: Same here, one assignment left to change

client/src/components/AddAssignmentsView.js
```diff
@@ -63,20 +73,21 @@ const AddAssignmentsView = () => {
       <Box borderRadius={1} sx={{ bgcolor: 'primary.light', p: '16px 12px', mb: 5, mt: 1, display: 'inline-block' }}>
         <Box sx={{ display: 'flex', flexWrap: 'wrap', justifyContent: 'space-between', alignItems: 'center' }}>
           <Typography align='left' sx={{ ml: 1.5 }} >Create and add a new study attainment:</Typography>
-          <Button variant='outlined' onClick={ () => navigate('/create-assignment/1') }>Create attainment</Button>
+          <Button variant='outlined' onClick={ () => navigate(`/${courseId}/create-temporary-attainment/${sisuInstanceId}`) }>Create attainment</Button>
```
kiiamakir, line 76: Could be divided into seperate lines to shorten line length

client/src/components/EditAssignmentView.js
```diff
@@ -3,41 +3,117 @@
...
+  };
+
+  const [attainments, setAttainments] = useState(getAttainment());
+
+  // Function to edit the data that is in the database
+  const editAttainment = async (attainmentObject) => {
+    try {
+      const attainment = await assignmentServices.editAttainment(courseId, instanceId, attainmentObject);
+      console.log(attainment);
+      //navigate('/' + courseId, { replace: true });
```
kiiamakir, line 52: This should be deleted if it's not used

client/src/components/EditAssignmentView.js
```diff
@@ -3,41 +3,117 @@
...
+      console.log(exception.message);
+    }
+  };
+
+  // Function to add data to the database
+  const addAttainment = async (attainmentObject) => {
+    try {
+      const attainment = await assignmentServices.addAttainment(courseId, instanceId, attainmentObject);
+      console.log(attainment);
+      //navigate('/' + courseId, { replace: true });
```
kiiamakir, line 63: Here as well

kiiamakir (changes requested):
Great work with this PR! Just a couple of code style nitpics

client/src/components/create-assignment/LeafAssignment.js
viivi-uhari: I kept Assignment here because the component name hasn't been changed.

client/src/components/create-assignment/ParentAssignment.js
viivi-uhari: Same here.

client/src/components/EditAssignmentView.js
viivi-uhari: I kept this here because I will uncomment this once the get-routes for attainments work. Also, possibly add a confirmation note for the user that adding/editing worked.

client/src/components/EditAssignmentView.js
viivi-uhari: Same here

kiiamakir (approved):

PR #174, opened 2023-03-11
Use Object.values for yup enum validation

no-comma:
Rather than manually listing all enum members in Yup validation schemas, which is prone to breakage when adding new enum members, `Object.values` returns all of the enum values automatically.


Gskartwii (approved):
Good catch!

jj-stigell (approved):

PR #175, opened 2023-03-11
Attainment deletion API

no-comma:
I've started to change parts of the code to just say "attainment" rather than "attainable" now, but not everything because that may interfere with other developers' work.

### DIFF
--- a/server/src/controllers/attainable.ts
+++ b/server/src/controllers/attainable.ts
@@ -164,7 +164,7 @@ export async function addAttainable(req: Request, res: Response): Promise<void> 
 export async function deleteAttainment(req: Request, res: Response): Promise<void> {
   /*
    * TODO: Check that the requester is logged in, 401 Unauthorized if not
-   * TODO: Check that the requester is authorized to delete attainables, 403
+   * TODO: Check that the requester is authorized to delete attainments, 403
    * Forbidden if not
    */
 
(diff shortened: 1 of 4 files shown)


server/src/controllers/attainable.ts
```diff
@@ -151,6 +161,70 @@ export async function addAttainable(req: Request, res: Response): Promise<void>
   });
 }
 
+export async function deleteAttainable(req: Request, res: Response): Promise<void> {
+  /*
+   * TODO: Check that the requester is logged in, 401 Unauthorized if not
+   * TODO: Check that the requester is authorized to add a course instance, 403
```
jj-stigell, line 167: This question does not only concern this PR but in general editing/deleting attainables.

Should each attainable have owner (reference to user's id) who is the only one allowed to edit/delete the attainable and its subattainables. Or should any user marked as teacher for the instance allowed to edit/delete any attainable on that instance? Or should editing be allowed for all teachers but deleting just for the attainable owner?

server/src/controllers/attainable.ts
```diff
@@ -18,6 +18,27 @@ import { findAttainableById, generateAttainableTag } from './utils/attainable';
 import { findCourseById } from './utils/course';
 import { findCourseInstanceById } from './utils/courseInstance';
 
+async function validateCourseAndInstance(
```
jj-stigell, line 21: Good refactoring!

server/src/controllers/attainable.ts
```diff
@@ -151,6 +161,70 @@ export async function addAttainable(req: Request, res: Response): Promise<void>
...
+      where: {
+        courseInstanceId: courseInstanceId
+      }
+    })
+  );
+
+  /*
+   * TODO: Also delete grades of these attainables.
+   * TODO: Query inside for loop may be bad, is there a better way to delete
+   * multiple entries?
```
jj-stigell, line 212: Since we have set the sequelize association for attainable parents as:
```javascript
Attainable.hasMany(Attainable, {
  onDelete: 'CASCADE',
  onUpdate: 'CASCADE'
});

Attainable.belongsTo(Attainable, {
  targetKey: 'id',
  foreignKey: 'attainableId'
});
```

The following should be enough for deleting the attainable and all of its subAttainables.
```javascript
await attainable.destroy();
```

jj-stigell (changes requested):

server/src/controllers/attainable.ts
no-comma: For a definitive answer we'll have to ask the PO, but from what I understand, attainables probably don't need any special access rights which can't be implemented with course instance roles. I would guess that teachers in charge have rights to edit, add, and delete all attainables of the instance, and other teachers can only add points.

server/src/controllers/attainable.ts
no-comma: Seems to work! Will this also delete the grades referencing these attainments ("attainables")?

jj-stigell: Yes, that associations is also set to cascade on delete 👍

jj-stigell (approved):
👍

Tonits (approved):
Looks good!

PR #177, opened 2023-03-12
Course view improvements

marillam:
Connected frontend course-view page to backend (only instances/courses, not yet formula or attainments). Added transitions to better show changes when instance is switched and made the instances table include all instances so that browsing is easier.

Refactored routes to ensure courseId is available and not confused to courseCode or sisuInstanceId.

New mock data file for tests, because the date comes as a string but I didn’t want to accidentally break anything.


client/src/App.js
```diff
@@ -100,20 +100,20 @@ function App() {
...
+              <Route path='/course-view/:courseId' element={<CourseView/>}/>
             </Route>
             { /* Pages that are only authorised for admin */ }
             <Route element={<PrivateRoute roles={[roles.admin]}/>}>
               <Route path='/create-course' element={<CreateCourseView/>}/>
             </Route>
             { /* Pages that are authorised for admin and teachers */ }
             <Route element={<PrivateRoute roles={[roles.admin, roles.teacher]}/>}>
-              <Route path='/fetch-instances/:courseId' element={<FetchInstancesView/>}/>
+              <Route path='/fetch-instances/:courseId/:courseCode' element={<FetchInstancesView/>}/>
```
kiiamakir, line 111: Should this be changed to :courseId/fetch-instances/:courseCode so the paths are consistent

client/src/tests/CourseView.test.js
```diff
@@ -3,72 +3,110 @@
...
-    expect(createInstanceButton).toBeDefined();
-    expect(addAssignmentButton).toBeDefined();
-    expect(seeAttendeesButton).toBeDefined();
+    await waitFor(() => {
+      const instanceInfo = getByText('Ongoing Instance');
+      const teachersInfo = getByText('Teachers in Charge');         
+      const assignments = getByText('Study Attainments');
+      const exercises = getByText('Exercises');  // Assignment
+      const projects = getByText('Project');  // Assignment
+      const exams = getAllByText('Exam');  // Assignmnets and instance types
```
kiiamakir, line 57: Attainment instead of assignment :)

kiiamakir (changes requested):
Great work, love the animation!

client/src/App.js
viivi-uhari: I actually did this change `':courseId/fetch-instances/:courseCode'` in #173 so if this is wanted for this branch as well, examples can be looked from there.

client/src/components/front-page/CourseTableRow.js
```diff
@@ -17,11 +17,11 @@ const CourseTableRow = ({ course }) => {
     <TableRow
       key={id} 
       hover={true}
-      onClick={() => { navigate('/course-view/' + courseCode); }}
+      onClick={() => { navigate('/course-view/' + id); }}
```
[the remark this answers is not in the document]
viivi-uhari: I added this comment line in my PR #173 to App.js:

<Route path='/course-view/:courseCode' element={<CourseView/>}/>
{ /* Path above will be replaced with '/course-view/:courseId' */ }

I think this change could be done in that branch once this branch is merged to main since I have done other actions in that branch as well (see comment above) so that merging this branch to that would be easier.

client/src/components/course-view/OngoingInstanceInfo.js
```diff
@@ -11,10 +11,8 @@ import LightLabelBoldValue from '../typography/LightLabelBoldValue';
 import textFormatServices from '../../services/textFormat';
 import useAuth from '../../hooks/useAuth';
 
-const mockInstitution = 'Aalto University';
-
 const OngoingInstanceInfo = ({ info }) => {
```
viivi-uhari, line 14: Should the title here be changed from 'Ongoing Instance' to 'Instance Details' (or something similar) since now details are shown here that aren't only about the ongoing instance but instead you can choose previous instances?

client/src/components/CourseView.js
```diff
@@ -2,50 +2,95 @@
...
-          (auth.role == 'SYSADMIN' || auth.role == 'TEACHER') && 
-          <Assignments assignments={mockAssignmentsClient} formula={'Weighted Average'} instance={mockCurrentInstance} /> /* TODO: Retrieve real formula */
-        }
-      </Box>
-      <Typography variant='h4' align='left' sx={{ fontWeight: 'light', mt: 8, mb: 3 }}>Past Instances</Typography>
-      <InstancesTable data={mockPastInstances} />
+      {courseDetails && currentInstance && instances &&
+        <> 
+          <Box sx={{ display: 'flex', flexWrap: 'wrap', justifyContent: 'space-between', alignItems: 'center', mb: 4 }}>
+            <Typography variant='h3' sx={{ fontWeight: 'light' }}>{courseDetails.courseCode + ' – ' + courseDetails.name.en}</Typography>
```
viivi-uhari, line 67: We could look into how to display titles better, since with this format the longer titles look really weird with my screen width. As an example a title looks like this (Sci is on its own line): 

```
MS-A0102 – Differential and Integral Calculus 1
                   (SCI)
```

As a quick fix, in my opinion lining the heading to the left (`textAlign: 'left', py: 1`) would be better but if you don't think that, this doesn't need to be changed.

viivi-uhari (changes requested):
Overall looks really good!

client/src/components/CourseView.js
marillam: I actually noticed the same thing and agreed! In my infinite wisdom I made the changes to the other branch, so they will come to main when it is merged.

client/src/App.js
marillam: Well use your changes, I don't think the work should be duplicated

client/src/components/course-view/OngoingInstanceInfo.js
marillam: Agreed

kiiamakir (approved):

viivi-uhari (approved):

Issue #161, opened 2023-02-22
Course points import using CSV


no-comma:
As a teacher, I want to upload (or paste) a CSV to enter points on partial grade category for a particular course instance, to store the course progress for eventual grade calculation.

The first column of the file should be student number that identifies student. The CSV file may have multiple columns, each identifying one assignment category. Therefore, the first line of CSV file should have a heading, a short string label that helps to assign the points to right assignment. Single CSV file can have multiple columns, allowing import of several assignments at the same time, but it does not have to include all assignment categories within course instance, so the ("partial") results may be imported in multiple phases

DoD: There is an UI for importing points from file, and import succeeds with a file with multiple students. Tests and documentation exist. It would be good to test also with several hundred students

- [x] Frontend design for adding points via CSV @viivi-uhari 
- [x] Frontend implementation for adding points via CSV @marillam 
- [x] API for adding grade data through a CSV file @jj-stigell 
- [x] Functionality for recording student points for attainments @jj-stigell 
- [x] Database design for storing student's points for attainments @jj-stigell 
- [X] Meeting about storing assignment points in database @jj-stigell @Tonits @no-comma

marillam:
- Frontend implementation done without connection to backend

PR #182, opened 2023-03-27
Database design for storing student's points for attainments:

jj-stigell:
Related #161.

This PR:
- Sets all user attributes nullable, allowing point adding to non-registered students (users). Routes that require that values are not null, must make the check on controller level (e.g., register user). 
- Adds index to user table `studentId` for faster querying.

Small refactoring included:
- remove redundant AALTO_GRADES_TEST_ENVIRONMENT, NODE_ENV already has environment information.
- use "typeof loginResult" when validating loginResult in auth controller.

### DIFF
--- a/server/src/database/models/user.ts
+++ b/server/src/database/models/user.ts
@@ -14,9 +14,9 @@ export default class User extends Model<
 > {
   declare id: CreationOptional<number>;
   declare studentId: CreationOptional<string>;
-  declare name: string;
-  declare email: string;
-  declare password: string;
+  declare name: CreationOptional<string>;
+  declare email: CreationOptional<string>;
+  declare password: CreationOptional<string>;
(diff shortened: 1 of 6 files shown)


server/src/database/models/user.ts
```diff
@@ -14,9 +14,9 @@ export default class User extends Model<
 > {
   declare id: CreationOptional<number>;
   declare studentId: CreationOptional<string>;
```
no-comma, line 16: This could also be changed to `studentNumber` in this PR.

no-comma (approved):
Let's change `studentId` to `studentNumber` later.

Gskartwii (approved):
Nice thing to use the more standard `NODE_ENV`. I'll second renaming the database field to `studentNumber`, and besides, all LGTM.

PR #183, opened 2023-03-31
Functionality for recording student points for attainments

jj-stigell:
Related to #161

Takes in to account the request from teachers that all validation errors from parsing CSV are returned, instead of one by one. This required updating the `ApiError` class to have an optional array of error codes.

`CourseRole` and `UserAttainmentGrade` have now composite key as primary key.

New mock data was generated for user and role table.

CSV uploading is tested with large sets, biggest one around 200k attainment grades. 97% of the excecution time goes on upserting users attainment grade:
```javascript
await UserAttainmentGrade.bulkCreate(preparedBulkCreate, { updateOnDuplicate: ['points'] });
```
If the actual CSV files uploaded to this endpoint are this size, there might be need for some optimizing in the future.


server/src/types/grades.ts
```diff
@@ -0,0 +1,20 @@
...
+export interface Student {
+  id?: number,
+  studentNumber: string,
+  grades: Array<Grade>
+}
+
+export interface AttainmentGrade {
+  userId: number,
+  attainableId: number,
+  points: number
```
no-comma, line 19: A "grade" can be either points or, for instance, a final course grade. For the latter "points" is not an appropriate term, so this shouldn't be `points`. Maybe `grade`, unless you have a better idea.

server/src/types/grades.ts
```diff
@@ -0,0 +1,20 @@
+// SPDX-FileCopyrightText: 2023 The Aalto Grades Developers
+//
+// SPDX-License-Identifier: MIT
+
+export interface Grade {
+  attainmentId: number,
+  points: number
+}
+
+export interface Student {
```
no-comma, line 10: It seems that the point of this interface is to store info about a student and their grades, so its name should reflect that. Maybe `GradesOfStudent`?

server/src/types/grades.ts
```diff
@@ -0,0 +1,20 @@
+// SPDX-FileCopyrightText: 2023 The Aalto Grades Developers
+//
+// SPDX-License-Identifier: MIT
+
+export interface Grade {
```
no-comma, line 5: Could have a comment for why this is separate from `AttainmentGrade`, and maybe a more descriptive name of the purpose of this interface.

server/src/types/grades.ts
```diff
@@ -0,0 +1,20 @@
...
+  points: number
+}
+
+export interface Student {
+  id?: number,
+  studentNumber: string,
+  grades: Array<Grade>
+}
+
+export interface AttainmentGrade {
```
no-comma, line 16: If this is intended to match the `UserAttainmentGrade` database model, it should be named `UserAttainmentGradeData` to be consistent with our other type naming. Granted, that is quite long so I can understand why you might have chosen not to name it that way, but I would personally still prefer it since it would immediately tell me that this is supposed to represent something from our database.

server/test/controllers/grades.test.ts
```diff
@@ -1,11 +1,17 @@
+/* eslint-disable max-len */
```
no-comma, line 1: I think that if a linter rule is disabled, it should always have a comment explaining why it is disabled. If we disable rules all the time without reason then that kind of defeats the purpose of having a linter.

no-comma (changes requested):

server/src/controllers/grades.ts
```diff
@@ -4,27 +4,139 @@
...
+import { AttainmentGrade, Grade, Student } from '../types/grades';
 import { HttpCode } from '../types/httpCode';
 import { validateCourseAndInstance } from './utils/courseInstance';
 
+/**
+ * Parse and extract attainment id's from the CSV file header.
+ * Correct format: "StudentNo,C3I9A1,C3I9A2,C3I9A3,C3I9A4,C3I9A5..."
+ * @param {Array<string>} header - Header part of the CSV file.
+ * @returns {Array<number>} - Array containing the id's of attainments.
+ * @throws {ApiError} - If first column not "StudentNo" (case-insensitive)
```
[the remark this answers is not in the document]
jj-stigell: This uses "StudentNo" to match the requirements in frontend, I do agree that "StudentNumber" is more self-describing. I'll leave it as is for now.

server/src/controllers/grades.ts
```diff
@@ -58,28 +170,172 @@ export async function addGrades(req: Request, res: Response, next: NextFunction)
...
-       */
-      console.log('CSV:', csvData);
-
-      res.status(HttpCode.Ok).json({
-        success: true,
-        data: {}
-      });
-      return;
+    .on('error', next) // Stream causes uncaught exception, pass error manually to the errorHandler.
+    .on('end', async function (): Promise<void> {
```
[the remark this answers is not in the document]
jj-stigell: updated to arrow function

server/src/types/error.ts
```diff
@@ -4,12 +4,27 @@
...
+  public readonly multiError: Array<string> | undefined;
 
-  constructor(message: string, statusCode: HttpCode) {
+  /**
+   * Creates an instance of ApiError.
+   * @param {string} message - The error message.
+   * @param {HttpCode} statusCode - The HTTP status code associated with the error.
+   * @param {Array<string>} [multiError] - Optional array of multiple error messages.
+  */
+  constructor(message: string, statusCode: HttpCode, multiError?: Array<string>,) {
```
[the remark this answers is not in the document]
jj-stigell: Good point, updated.

server/test/controllers/sisu.test.ts
```diff
@@ -14,6 +14,25 @@ const mockedAxios: jest.Mocked<AxiosStatic> = axios as jest.Mocked<typeof axios>
 
 const request: supertest.SuperTest<supertest.Test> = supertest(app);
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function checkRes(courseInstance: any): void {
```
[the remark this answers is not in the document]
jj-stigell: updated to type `CourseInstanceData`

server/test/controllers/grades.test.ts
jj-stigell: removed lint disabling. had some test results with too long strings, fixed the length issue.

server/src/types/grades.ts
jj-stigell: updated to `UserAttainmentGradeData`

server/src/controllers/grades.ts
```diff
@@ -58,28 +170,172 @@ export async function addGrades(req: Request, res: Response, next: NextFunction)
...
+          if (nonExistingStudents.length > 0) {
+            const newUsers: Array<User> = await User.bulkCreate(
+              nonExistingStudents.map((studentNo: string) => {
+                return {
+                  studentId: studentNo
+                };
+              }), { transaction: t }
+            );
+
+            await CourseInstanceRole.bulkCreate(
```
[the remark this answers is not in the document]
jj-stigell: Yes, good point! updated

server/src/types/grades.ts
jj-stigell: I am using `points` as the database model uses `points` instead of `grades`.

server/src/types/grades.ts
jj-stigell: updated to `StudentGrades`

server/src/types/grades.ts
jj-stigell: Actually after thinking about this, we don't need the `Grade` interface. Refactored so that only `AttainmentGrade` used.

server/src/controllers/grades.ts
```diff
@@ -4,27 +4,139 @@
...
+ * an array of their grades.
+ * @throws {ApiError} - If there is an error in the CSV file (e.g. incorrect data type in a cell).
+ * Collects all errors found to an array, does not throw error immediately on first incorrect value.
+*/
+export function parseGrades(
+  studentGradingData: Array<Array<string>>, attainmentIds: Array<number>
+): Array<Student> {
+  const students: Array<Student> = [];
+  const errors: Array<string> = [];
+  let currentRow: number = 2; // Takes into consideration header row and start index of 0.
```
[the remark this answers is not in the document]
jj-stigell: updated

server/src/controllers/grades.ts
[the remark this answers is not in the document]
jj-stigell: Or would it be better not to check the first column at all (it could be anything, e.g, "studentNo", "studentNumber", "-", etc.). We don't really use the "StudentNo" (row 1, col 1) for anything in the back end. This header 1 row 1 col check in the backend was more of a just to enforce strict style to the CSV teachers upload.

server/src/types/grades.ts
no-comma: Then I think the database model should also be updated, although that's not critical for this PR, it can be changed later.

server/src/controllers/grades.ts
```diff
@@ -58,28 +181,164 @@ export async function addGrades(req: Request, res: Response, next: NextFunction)
...
+                };
+              });
+            return studentGradingData;
+          });
+
+        // TODO: Optimize if datasets are big.
+        await UserAttainmentGrade.bulkCreate(preparedBulkCreate, { updateOnDuplicate: ['points'] });
+
+        // After this point all students attainemnt grades are created or updated
+        // in the database for each student.
```
no-comma, line 332:
```suggestion
        // After this point all the students' attainment grades have been created or
        // updated in the database.
```

server/src/middleware/errorHandler.ts
```diff
@@ -42,6 +42,15 @@ export function errorHandler(err: unknown, req: Request, res: Response, next: Ne
   if (err instanceof ApiError) {
     res.status(err.statusCode);
 
+    // If multiple errors not undefined, return that in errors property.
```
no-comma, line 45: We could have a `getErrors(): Array<string>` method in `ApiError`, which returns either a single error message in an array of length 1, or the `multipleErrors` property, so we don't need to check whether `multipleErrors` is defined here.

server/src/controllers/grades.ts
```diff
@@ -4,27 +4,150 @@
...
+   * Both of these values are incremented after row/column parsing operation.
+   *
+   *        | column 1  | column 2 | column 3 |
+   *  --------------------------------------------
+   *  row 1:| StudentNo | C1I1A1   | C1I1A6   |
+   *  row 2:| 812472    | 12       | 32       |
+   *  row 3:| 545761    | 0        | 15       |
+   *  ...
+   *  row n:| ...
+   */
```
no-comma, line 93:
```suggestion
  /**
   * currentRow and currentColumn are user facing row and column numbers of the
   * uploaded CSV file. They are the index of the row and column plus 1,
   * so the first row and column in the CSV file will have the index 0 but the
   * number 1. See an example of row and column numbers below.
   *
   *        | column 1  | column 2 | column 3 |
   *  --------------------------------------------
   *  row 1:| StudentNo | C1I1A1   | C1I1A6   |
   *  row 2:| 812472    | 12       | 32       |
   *  row 3:| 545761    | 0        | 15       |
   *  ...
   *  row n:| ...
   *
   * currentRow is incremented after a row has been parsed, currentColumn is
   * incremented after a column of a row has been parsed and is reset to 2
   * when the current row has been fully parsed.
   *
   * Row and column 1 are handled separately.
   */
```

no-comma (changes requested):
The code looks good and is well commented. Some clarifications should still be included and potentially a minor change to `ApiError`, once these are addressed I'll approve this.

server/src/types/grades.ts
jj-stigell: Agree, but better to handle this in another PR.

server/src/middleware/errorHandler.ts
jj-stigell: Good point, added getter to the ApiError class and updated error handler.

no-comma (approved):

PR #185, opened 2023-03-31
Course results view

kiiamakir:
Not connected to backend. This PR contains the table for viewing grades as well as buttons for importing grades to the table and calculating final grades


client/src/components/course-results-view/CourseResultsTableHead.js
```diff
@@ -0,0 +1,71 @@
...
+          </TableCell>
+        ))}
+      </TableRow>
+    </TableHead>
+  );
+};
+  
+CourseResultsTableHead.propTypes = {
+  onRequestSort: PropTypes.func.isRequired,
+  order: PropTypes.oneOf(['asc', 'desc']).isRequired,
```
marillam, line 66: nice!

client/src/components/course-results-view/CourseResultsTable.js
```diff
@@ -0,0 +1,198 @@
...
+  const stabilizedThis = array.map((el, index) => [el, index]);
+  stabilizedThis.sort((a, b) => {
+    const order = comparator(a[0], b[0]);
+    if (order !== 0) {
+      return order;
+    }
+    return a[1] - b[1];
+  });
+  return stabilizedThis.map((el) => el[0]);
+};
```
marillam, line 48: I wonder if this should be in a service (sorting)? Is it generic? It will probably need a few comments but it would make this component lighter too if these logic wasn't here. But only if it's an easy change!

marillam (changes requested):
Great job overall! The table is really pretty and everything seems to work well :)

client/src/components/course-results-view/CourseResultsTable.js
```diff
@@ -0,0 +1,198 @@
...
+            rowsPerPage={rowsPerPage}
+            page={page}
+            onPageChange={handleChangePage}
+            onRowsPerPageChange={handleChangeRowsPerPage}
+          />
+        </Box>
+      </Paper>
+      <FormControlLabel
+        control={<Switch checked={dense} onChange={handleChangeDense} />}
+        label="Dense padding"
```
viivi-uhari, line 185: Love this!

viivi-uhari (changes requested):
Great job, the design and functionality are amazing. And the code is so clean! I almost have nothing to add.

Could the buttons be changed to different locations since the See Course Results button feels to get a bit lost. I think a quick and a nice fix would be to move the 'Add Attainment' button to the top as in Figma. But this probably doesn't need to be changed in this branch and we can discuss it during the next sprint.

Also, now the calculation of final grades is possible even if you haven't imported any grades but I guess this is only a feature of the mock data so nothing needs to be changed.

client/src/components/course-results-view/CourseResultsTable.js
kiiamakir: Modified the final grade calculation so that it sends an error message if no student grades have been added -> now the mock data also works as intended

viivi-uhari (approved):
Great job!

marillam (approved):

PR #189, opened 2023-04-11
Delete attainment functionality in frontend

viivi-uhari:
Functionality for deleting attainments from the DB via frontend. When an attainment is deleted all its sub-attainments are also deleted.


kiiamakir (approved):
Looks good! Testing was a bit difficult since the get attainments route hasn't been implemented. We should get this to main before the formula attainment refactor PR so that the refactor branch can be modified accordingly.

client/src/components/EditAssignmentView.js
```diff
@@ -66,6 +67,21 @@ const EditAssignmentView = () => {
     }
   };
 
+  // Function to delete data from the database or from the context
+  const deleteAttainment = async (attainmentId) => {
+    // If this view is opened from the course view, delete from DB
+    // Else the attainment is being edited during the creation of an instance so only delete from the context
+    if (instanceId) {
```
marillam, line 74: Should this be inside try/catch? Or is error handling done somewhere else?

marillam (changes requested):
Looks good for the most part! Waiting for the get attainment api to work so this is easier to test in the UI but this is a good start

client/src/components/EditAssignmentView.js
viivi-uhari: Added try/catch

marillam (approved):

PR #190, opened 2023-04-14
Connect csv to backend

marillam:
Connects import grades from file (css upload) dialog to backend.

- upload function checks response and in case of failure the dialog shows errors from server to the user
- added snack pack to inform the user about success/failure
- created own test file for the component and refactored one test from course view to it

@kiiamakir I didn't change anything about the returnImportedGrades. If something needs to be done, suggest in the review.

### DIFF
--- a/client/src/tests/CourseView.test.js
+++ b/client/src/tests/CourseView.test.js
@@ -109,31 +109,4 @@ describe('Tests for CourseView component', () => {
     expect(firstTeacherInCharge).not.toBeInTheDocument();
   });
 
-  test('CourseView should show a dialog for uploading a file when clicking on a menu button and choosing that option', async () => {
-
-    const auth = { role: 'TEACHER' };
-    const { getByText, findByText } = renderCourseView(auth);
-
-    const importGradesMenuButton = await findByText('Import grades');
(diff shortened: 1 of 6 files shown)


client/src/tests/CourseView.test.js
```diff
@@ -107,31 +107,4 @@ describe('Tests for CourseView component', () => {
     expect(firstTeacherInCharge).not.toBeInTheDocument();
   });
 
-  test('CourseView should show a dialog for uploading a file when clicking on a menu button and choosing that option', async () => {
```
viivi-uhari, line 110: Why is this test deleted?

viivi-uhari (changes requested):
Good job. The CSV import seems to work, but there's almost no checking of the file implemented in backend yet and backend isn't connected to DB yet so, changes done there might affect frontend, but for now this is a good start!

client/src/tests/CourseView.test.js
kiiamakir: I tink this test was moved to the FileLoadDialog.test file

marillam: Yep, it was just moved.

kiiamakir (approved):

PR #195, opened 2023-04-16
Frontend refactor of formula attributes

kiiamakir:
Frontend code refactored so that attainments now list the formula attributes as objects instead of an array. This was done to match the implementation in backend. In order for the code to work, the backend should return a list of attributes that match the keys in the attainments' attribute object.

An example:

An attainment that uses the formula x lists the attributes as an object

formulaAttributes: {
maxPoints: "10",
minRequiredPoints: "5",
weight: "0.5"
}

Formula x includes the same keys in an array

["maxPoints", "minRequiredPoints", "weight"]

### DIFF
--- a/client/src/components/create-assignment/LeafAssignment.js
+++ b/client/src/components/create-assignment/LeafAssignment.js
@@ -14,6 +14,7 @@ import ConfirmationDialog from './ConfirmationDialog';
 import StringTextField from './StringTextField';
 import DateTextField from './DateTextField';
 import assignmentServices from '../../services/assignments';
+import formulaService from '../../services/formulas';
 
 // An Assignmnet component without subAttainments and hence without a formula as well.
 // If this isn't the root Assignment, this can be deleted
@@ -45,11 +46,12 @@ const expiryData = {
 */
 const AttributeTextFields = ({ formulaAttributeNames, indices, setAttainments, attainments }) => {
   return (
-    formulaAttributeNames.map((attribute, index) => {
+    formulaAttributeNames.map((attribute) => {
+      const attributeLabel = formulaService.getAttributeLabel(attribute);
       return(
         <StringTextField 
-          key={index}
-          fieldData={{ fieldId: 'attribute' + index, fieldLabel: attribute }}
+          key={attribute}
+          fieldData={{ fieldId: 'attribute_'+attribute, fieldLabel: attributeLabel }}
           indices={indices} 
           setAttainments={setAttainments} 
           attainments={attainments}
--- a/client/src/components/create-assignment/LeafAssignment.js
+++ b/client/src/components/create-assignment/LeafAssignment.js
@@ -51,7 +51,7 @@ const AttributeTextFields = ({ formulaAttributeNames, indices, setAttainments, a
       return(
         <StringTextField 
           key={attribute}
-          fieldData={{ fieldId: 'attribute_'+attribute, fieldLabel: attributeLabel }}
+          fieldData={{ fieldId: 'attribute_' + attribute, fieldLabel: attributeLabel }}
           indices={indices} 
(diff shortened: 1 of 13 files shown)


client/src/components/create-assignment/LeafAssignment.js
```diff
@@ -45,11 +46,12 @@ const expiryData = {
...
   return (
-    formulaAttributeNames.map((attribute, index) => {
+    formulaAttributeNames.map((attribute) => {
+      const attributeLabel = formulaService.getAttributeLabel(attribute);
       return(
         <StringTextField 
-          key={index}
-          fieldData={{ fieldId: 'attribute' + index, fieldLabel: attribute }}
+          key={attribute}
+          fieldData={{ fieldId: 'attribute_'+attribute, fieldLabel: attributeLabel }}
```
viivi-uhari, line 54: I prefer spaces around operators (+). Could this be changed?

client/src/components/create-assignment/ParentAssignment.js
```diff
@@ -52,6 +52,7 @@ const ParentAssignment = ({ indices, addSubAttainments, setAttainments, attainme
         <Typography variant="body1" sx={{ flexGrow: 1, textAlign: 'left', mb: 0.5 }}>
           {'Grading Formula: ' + formulaName}
         </Typography>
+        { /* Navigation below doesn't work because formula selection has only been implemented for course grade */ }
```
viivi-uhari, line 55: Good clarification 👍

client/src/mock-data/mockAttainmentsClient.js
```diff
@@ -17,15 +17,23 @@ const mockAttainmentsClient = [
         expiryDate: new Date(2024, 8, 14), 
         formulaId: 1,
         affectCalculation: true,
-        formulaAttributes: ['', '', ''],
+        formulaAttributes: {
```
viivi-uhari, line 20: Nice refactor, hopefully backend complies with this.

client/src/services/formulas.js
```diff
@@ -33,4 +33,13 @@ const getFormulaAttributes = (formulaId) => {
   return formulaAttributes;
 };
 
-export default { getFormulas, setFormula, getFormulaName, getFormulaAttributes };
+// A function for getting the displayable label from a label key
+// for example: the label key maxPoints would return Max Points
+const getAttributeLabel = (labelKey) => {
```
viivi-uhari, line 38: Great that you got this function working!

viivi-uhari (changes requested):
Well done (only a small change could be done)! Everything works and backend should be able to comply with this changes since this is what we agreed on in the meeting with the backend team.

client/src/components/create-assignment/LeafAssignment.js
```diff
@@ -45,11 +46,12 @@ const expiryData = {
 */
```
viivi-uhari, line 46: I also noticed lines where spaces aren't used inside curly brackets in this file, so if you don't mind then that could be corrected. Lines 171 (`{JSON.stringify(indices) !== '[0]' ?`) and 176 (`<Box sx={{ width: '1px' }}/>}`).

viivi-uhari (approved):

viivi-uhari:
I converted this PR to a draft issue so that it doesn't merge to main before #189 is merged.

client/src/services/formulas.js
marillam, line 38: Not a mandatory change!! But I wonder if this should be in text format services or something since it's not specific to just formulas even if it is used just here.

marillam (commented):
Looks great! Very clean code and the refactor seems thorough

marillam (approved):

PR #196, opened 2023-04-20
Deployment configuration

lakooi:


.github/workflows/main-merge-workflow.yml
no-comma, line 1: Rename to `.yaml` for consistency.

docker-compose-with-images.yaml
```diff
@@ -0,0 +1,54 @@
+# SPDX-FileCopyrightText: 2023 The Aalto Grades Developers
+#
+# SPDX-License-Identifier: MIT
+
```
no-comma, line 4: Add a comment explaining the purpose of this Docker Compose configuration, similar to other Docker Compose files.

docker-compose-with-images.yaml
```diff
@@ -0,0 +1,54 @@
...
+    networks:
+      - local
+
+  database:
+    image: postgres
+    restart: always
+    environment:
+      POSTGRES_PASSWORD: "${POSTGRES_PASSWORD}"
+    networks:
+      - local
```
no-comma, line 50: This is identical to the common database service, so it should be extended instead:
```suggestion
  database:
    extends:
      file: ./common-services.yaml
      service: database
```

docker-compose-with-images.yaml
```diff
@@ -0,0 +1,54 @@
...
+    container_name: backend-server
+    environment:
+      POSTGRES_URL: "database"
+      POSTGRES_USER: "postgres"
+      POSTGRES_DATABASE: "postgres"
+      AALTO_GRADES_JWT_SECRET: "${AALTO_GRADES_JWT_SECRET}"
+      POSTGRES_PASSWORD: "${POSTGRES_PASSWORD}"
+      SISU_API_KEY: "${SISU_API_KEY}"
+      SISU_API_URL: "${SISU_API_URL}"
+      AALTO_GRADES_FRONTEND_CORS_ORIGIN: "https://https://aalto-grades.cs.aalto.fi"
```
no-comma, line 40: This should be changed to extend a common service since it's redefining so many environment variables, but it's fine for now, just change the order a bit:
```suggestion
      POSTGRES_URL: "database"
      POSTGRES_USER: "postgres"
      POSTGRES_DATABASE: "postgres"
      POSTGRES_PASSWORD: "${POSTGRES_PASSWORD}"
      SISU_API_KEY: "${SISU_API_KEY}"
      SISU_API_URL: "${SISU_API_URL}"
      AALTO_GRADES_JWT_SECRET: "${AALTO_GRADES_JWT_SECRET}"
      AALTO_GRADES_FRONTEND_CORS_ORIGIN: "https://https://aalto-grades.cs.aalto.fi"
```

no-comma (changes requested):

common-services.yaml
no-comma, line 1: The "standard" frontend and backend containers should extend the minimum versions.

no-comma (changes requested):
My previous unaddressed comments still apply, other than those just one more detail and then I'll approve this.

common-services.yaml
lakooi: Do you mean you would like the minimum to contain the build step and the other one to not have it?

lakooi: If so, I would rename the currently existing frontend-minimum and backend-minimum to for example frotend-no-build and backend-no-build

no-comma: The only difference between `frontend` and `frontend-minimum` as well as `backend` and `backend-minimum` is that the non-minimum versions contain build steps. Since everything else is identical, `frontend` should extend `frontend-minimum` with the build step, and same for `backend`.

no-comma (approved):
LGTM

jj-stigell (approved):

PR #199, opened 2023-05-05
Backend dev script for local development

jj-stigell:
Compile and restart backend automatically after changes to files.


PasiSa (approved):
Looks ok.

Issue #228, opened 2023-05-23
Course role access control


PasiSa:
- [x] create mapping for Course specific "teacher-in-charge" role mapping courses and users
- [x] access control for the above
- [ ] access control for "teacher" role mapping instances and users (tables already exist)

Backend (old text):

- Course-specific roles ['STUDENT', 'TEACHER', 'TEACHER_IN_CHARGE'] are already implemented.

- Adding system level roles 'ADMIN' and 'USER', requires new attribute 'ROLE' to user table 

Frontend:

- Frontend renders components conditionally based on the role returned in login/self-info. Currently following system level roles are used in frontend ['SYSADMIN', 'TEACHER', 'STUDENT', 'ASSISTANT']

- Updating roles require refactoring, conditional rendering must be updated to handle changes. Create new course view will be available to only role 'ADMIN', all other views available to both 'ADMIN' and 'USER'.

- On course level, conditional rendering on based on course-specific role

Related issues:
- [x] https://github.com/aalto-grades/base-repository/issues/240
- [x] https://github.com/aalto-grades/base-repository/issues/241

PR #231, opened 2023-05-24
Update authorization to API endpoints

jj-stigell:
Authorization with JWT required on all API endpoint (except signup and login).

### DIFF
--- a/server/src/middleware/authorization.ts
+++ b/server/src/middleware/authorization.ts
@@ -1,0 +1,22 @@
+// SPDX-FileCopyrightText: 2023 The Aalto Grades Developers
+//
+// SPDX-License-Identifier: MIT
+
+import { NextFunction, Request, Response } from 'express';
+import passport from 'passport';
+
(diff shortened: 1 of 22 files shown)


server/src/middleware/authorization.ts
```diff
@@ -0,0 +1,22 @@
...
+//
+// SPDX-License-Identifier: MIT
+
+import { NextFunction, Request, Response } from 'express';
+import passport from 'passport';
+
+import { JwtClaims } from '../controllers/auth';
+import { HttpCode } from '../types/httpCode';
+
+export function authorization(req: Request, res: Response, next: NextFunction): void {
```
no-comma, line 11: Shouldn't this be `authentication`? Or is this intended to also check for authorization in the future?

jj-stigell: Yes, I implemented this with authorization (role checking from token) in mind. We can extend this to handle the "system level" role check, if we implement it as planned in #228.

Issue #238, opened 2023-06-01
Make front-end to use Typescript


PasiSa:
Frontend code should be modified to use TypeScript, with eventual steps so that nothing that works now will not be broken.

- [x] Initial change #249 
- [x] #259 
- [x] tsconfig strict set to true
- [x] Enforce the same ESLint rules as the backend

PR #249, opened 2023-06-05
Frontend TypeScript

no-comma:
This PR is intended to start moving the frontend code to TypeScript to a level where it runs normally as before, so type definitions can easily be added later as we go along.

Related to: #238


client/src/components/course-results-view/SisuExportDialog.tsx
```diff
@@ -159,22 +161,29 @@ const SisuExportDialog = ({ open, handleClose }) => {
                 InputLabelProps={{ shrink: true }}
                 type="date"
                 label="Assessment Date"
-                format="DD-MM-YYYY"
+                /* TODO: Fix TS */
+                //format="DD-MM-YYYY"
```
no-comma, line 165: `format` doesn't appear to be a property of `TextField` at all: https://mui.com/toolpad/reference/components/text-field/

This doesn't seem to format the date entry from what I tried even in the JavaScript version, so it seems to me that this is just an uncaught error. If so, these lines should just be removed and we can have another PR where we fix all of the date entries, there seem to be others that are also formatted as MM-DD-YYYY.

jj-stigell (approved):
LGTM

Issue #241, opened 2023-06-01
Backend support for Admin actions


PasiSa:
Admin users should be able to access all courses regardless of the teacher roles.
Some actions, such as creating courses and defining teachers should only be available for admin users.

no-comma:
Initially, let's implement it so that admin's have all rights and regular users have practically no rights.

Issue #229, opened 2023-05-23
Implement JWT authentication on API endpoints


PasiSa:
- [x] Identify endpoints that require authentication -> all endpoints except signup and login
- [x] Check auth on router level, valid JWT token must be provided in cookie #231 
- [x] Modify controllers to include checks for authorization on system level where necessary (e.g., only admins can create courses)
- [x] Add tests to cover the changes

JWT stores users system level role

PR #251, opened 2023-06-05
System level auth

jj-stigell:
Backend implementation for role based authorization.

Routes that need to be protected with role access rights -> include the authorization middleware. Allowed roles must be defined for the route as an array `Array<UserRole>`.

Following route(s) require role 'ADMIN' in the jwt token payload
- Creating a new course

Any other routes we should protect?

related to #241 and #229


no-comma:
I'll review this on Wednesday.

server/test/controllers/auth.test.ts
```diff
@@ -60,7 +60,7 @@ describe('Test POST /v1/auth/signup - create a new user', () => {
         name: 'aalto',
         password: 'grades',
         studentID: '123456',
-        role: 'SYSADMIN'
+        role: 'ADMIN'
```
no-comma, line 63: Use the `enum` instead? Also applies to the other `'ADMIN'` change in this file.

no-comma (changes requested):
I had a chance to look at this now already, and other than one minor detail everything looks good.

client/src/types/general.ts
```diff
@@ -0,0 +1,9 @@
+// SPDX-FileCopyrightText: 2023 The Aalto Grades Developers
+//
+// SPDX-License-Identifier: MIT
+
+/* eslint-disable no-unused-vars */
+export enum UserRole {
+  User = 'USER',
+  Admin = 'ADMIN'
+}
```
no-comma: This works for this PR, but I would like to have a single code module for types shared by the frontend and backend, so we don't need to manually keep them in sync or duplicate code between `server` and `client`.

jj-stigell: So something like folder "common" to the project root and subfolder "types" where we place client/server shared types?

jj-stigell: So somehow with declaring global types and no need for importing. I will leave this for now as is, let's think this one later.

client/src/components/FrontPage.tsx
```diff
@@ -31,9 +35,16 @@ const FrontPage = (): JSX.Element => {
...
           Your Current Courses
         </Typography>
+        { /* admins are shown the button for creating a new course */
+          auth.role == SystemRole.Admin &&
+          <Button id='ag_new_course_btn' size='large' variant='contained' onClick={() => {
+            navigate('/create-course');
+          }}>
+            Create New Course
+          </Button>
+        }
```
no-comma, line 45: Courses are created from `AllCoursesView` as of #254, so this should be removed.

client/src/tests/FrontPage.test.tsx
```diff
@@ -51,4 +52,20 @@ describe('Tests for FrontPage component', () => {
...
+
+
+  test('FrontPage should render Create New Course for admin', async () => {
+
+    const auth = { role: SystemRole.Admin };
+    renderFrontPage(auth);
+    await waitFor(() => expect(screen.queryByText('Create New Course')).toBeInTheDocument());
+
+  });
+
```
no-comma, line 70: Same here.

no-comma (changes requested):

no-comma (approved):
LGTM

PR #254, opened 2023-06-07
Admin front page

no-comma:
This PR adds a view showing all existing courses, which is to be used as the front page of admin users.


jj-stigell (approved):
LGTM. Once the amount of courses increases there might be need to incorporate pagination and filter/search functionality to the admin front page

PR #258, opened 2023-06-07
Redirect to /course-view after creating a course

no-comma:
After creating a course, the user would be redirected to the front page at `/`, but now that courses are instead created from `/course-view`, the user should be redirected back to `/course-view` after creating a course.

### DIFF
--- a/client/src/components/CreateCourseView.tsx
+++ b/client/src/components/CreateCourseView.tsx
@@ -15,7 +15,7 @@ const CreateCourseView = () => {
     try {
       const course = await coursesService.addCourse(courseObject);
       console.log(course);
-      navigate('/', { replace: true });
+      navigate('/course-view', { replace: true });
     } catch (exception) {
       console.log(exception.message);
     }


client/src/components/CreateCourseView.tsx
```diff
@@ -15,7 +15,7 @@ const CreateCourseView = () => {
     try {
       const course = await coursesService.addCourse(courseObject);
       console.log(course);
-      navigate('/', { replace: true });
+      navigate('/course-view', { replace: true });
```
jj-stigell, line 18: Would it make more sense if we redirect to the newly created course page?

jj-stigell (changes requested):

client/src/components/CreateCourseView.tsx
no-comma: Yes, that would probably be better, but it will require some other changes as well because currently the newly created course's page will be empty since it has no instances. I could make a change that allows the user to see the option to create a new instance even with no instances currently.

jj-stigell: I think its better to have redirection to the course page even if its empty. If not this could become frustrating once there are hundreds of courses and admin has to navigate and find the newly created course from main course view, if they want for example add more teachers to the course, etc.

Issue #246, opened 2023-06-02
Front-end should fetch results data from backend


PasiSa:
Front-end still uses some mock data locally, and does not fetch the data using APIs. Some of the API endpoints may need to be modified.

- [x] #205 
- [x] Remove all non-test mock data from frontend
- [x] #282

PR #264, opened 2023-06-09
Store final grades to db

jj-stigell:
After grade calculation, save the grades to database instead of just sending the results back to the user.

related to #246


server/src/controllers/grades.ts
```diff
@@ -611,19 +590,19 @@ export async function calculateGrades(
    * same information as unorganizedPresetGrades.
    * Student number -> Formula node -> Preset grade.
```
no-comma, line 591:
```suggestion
   * User ID -> Formula node -> Preset grade.
```

no-comma (approved):
Just one comment update which you can quickly commit and merge, otherwise looks good.

In the future, we may want to save all the intermediate grades as well, e.g., the grade of an exam and exercises, rather than just the final grade.

Issue #222, opened 2023-05-23
Instance creation progress tracking


jj-stigell:
Instance creation progress could be visualised with for example [stepper](https://mui.com/material-ui/react-stepper/)

no-comma:
Course instance creation no longer has multiple steps.

PR #266, opened 2023-06-09
Allow cancelling course instance creation

no-comma:
Implements part of #222 

Adds an option to cancel course instance creation in the basic instance information editing page and Sisu instance fetch page.


jj-stigell (approved):
Looks good. At some point I had in mind it would be nice/informative to have something like [MUI stepper](https://mui.com/material-ui/react-stepper/) that would show the progress and allow backtrack between tabs during instance creation.

no-comma:
> Looks good. At some point I had in mind it would be nice/informative to have something like [MUI stepper](https://mui.com/material-ui/react-stepper/) that would show the progress and allow backtrack between tabs during instance creation.

That could still be done, but in another PR for now.

PR #267, opened 2023-06-11
Get single attainment

jyrge:
API functionality for getting an attainment, optionally with the tree, as described in #138.

Additionally, replace all occurrences of "attainable" with "attainment".


no-comma (approved):
LGTM

PR #276, opened 2023-06-13
CSV new tags

### LINKED ISSUE #270: Update CSV format for new attainment tags
The CSV imports still use the old generated tag format, but now tags are manually definable so the CSV format should be updated to match that.

no-comma:
Closes #270


jj-stigell (approved):
Nice work. This custom tagging looks much better than the previous attainment id based tags.

PR #278, opened 2023-06-13
Get final grades from server

jj-stigell:
After calculating the final grades, get grading data from server. Deletes grading mock data from frontend.


client/src/components/CourseResultsView.tsx
```diff
@@ -40,43 +47,45 @@ const CourseResultsView = () => {
...
-    if (students.length === 0) {
+    try {
       setSnackPack((prev) => [...prev,
-        { msg: 'Import student grades before calculating the final grade.', severity: 'error' }
+        { msg: 'Calculating final grades...', severity: 'info' }
       ]);
-    } else {
-      try {
-        // TODO: connect to backend
+      await sleep(2000);
```
no-comma, line 58: What is the purpose of these `sleep` calls?

no-comma (approved):
Looks good, just one question.

client/src/components/CourseResultsView.tsx
jj-stigell: I think the original idea was that the previous snackbar would not be replaced so quickly with other (user has time to read before text changes). In the future this should be updated so that the snackbar can be stacked instead of replaced so there would be no need for this kind of sleep timer.

PR #314, opened 2023-06-27
Update front page

### LINKED ISSUE #311: Remove distinction between "previous" and "current" courses in front page
With #288 the current front page layout may need to be updated.

- [ ] Update front page
- [ ] Update courses of user API endpoint and tests

no-comma:
Closes #311


client/src/components/FrontPage.tsx
```diff
@@ -31,15 +41,42 @@ function FrontPage(): JSX.Element {
...
+            Create New Course
+          </Button>
+        }
       </Box>
-      <BasicGrid data={currentCourses} />
-      <Typography variant="h2" align="left" sx={{ flexGrow: 1, mt: 4 }}>
-        Inactive Courses
-      </Typography>
-      <CourseTable data={previousCourses} />
+      <CourseTable data={courses} />
```
jj-stigell, line 79: Should the "Courses" under "Your courses" render conditionally only for admins?

Now non admin can see empty "Courses" view.

jj-stigell (commented):
one comment

client/src/components/FrontPage.tsx
no-comma: I decided to remove `AllCoursesView` and integrate the list to `FrontPage` in this PR because there is very little to differentiate the two views now. But I think it's probably useful in some cases for regular users to see a list of all courses as well (and there's no reason to hide that anyway), so I chose not to conditionally render it.

The list being empty was unintentional, I didn't realize the use of `GET /v1/courses` was locked behind the admin role before.

jj-stigell (approved):
👍

PR #318, opened 2023-06-28
Show tags in the course view attainment list

jj-stigell:


no-comma (commented):
Looks fine otherwise, but I would remove the "tag: " prefix. The tag is basically another name for an attainment and I certainly wouldn't use a "name: " prefix either.

no-comma (approved):

PR #324, opened 2023-06-30
Update/remove mockdata from course creation view

jj-stigell:
* Remove mock data from course creation view
* Use formik for handling form validation with yup
* Enable snackpack/alerts to map multiple errors into view
* Add all missing fields for course creation
* Responsible teachers are inputted by their email addresses, change updated to backend course creation API

related to #246


server/src/routes/course.ts
```diff
@@ -51,6 +51,14 @@ export const router: Router = Router();
  *     type: integer
  *     description: Maximum amount credits student can receive from passing the course.
  *     example: 5
+ *   teachersInCharge:
```
no-comma, line 54:
```suggestion
 *   TeachersInCharge:
```

server/src/routes/course.ts
```diff
@@ -178,6 +186,8 @@ router.get(
  *                 $ref: '#/definitions/LocalizedString'
  *               name:
  *                 $ref: '#/definitions/LocalizedString'
+ *               teachersInCharge:
+ *                 $ref: '#/definitions/teachersInCharge'
```
no-comma, line 190:
```suggestion
 *                 $ref: '#/definitions/TeachersInCharge'
```

client/src/types/index.tsx
```diff
@@ -15,8 +16,14 @@ export type FullResponse<T> = AxiosResponse<ApiResponse<T>, unknown>;
...
   severity?: 'error' | 'warning' | 'info' | 'success'
 }
 
 export type State<T> = [T, (value: T) => void];
+
+export interface NewCourseData extends Omit<
+CourseData, 'teachersInCharge' | 'id' | 'evaluationInformation'
+> {
+  teachersInCharge: Array<string>;
+}
```
no-comma, line 29: I think it would be cleaner if we avoid adding new types for POST request data unless it is somehow significantly different from the corresponding GET request data. In this case, I would add an optional `email` field to `UserData` and make `evaluationInformation` optional (or completely remove it since it has remained unused ever since it was added) in `CourseData`, and then we could just use that.

no-comma (commented):
Looks good, mainly just a comment on types.

client/src/types/index.tsx
jj-stigell: Good point, remove the new type and updated old one. Kept `evaluationInformation` as is (set as optional).

no-comma (approved):

PR #331, opened 2023-07-04
UI updates/additions and get only relevant course instances for the course view.

### LINKED ISSUE #320: Getting course instances returns every instance regardless of course
The `GET /v1/courses/:courseId/instances` API endpoint is intended to return every instance of the given course, but instead it returns every course instance in the database regardless of which course it actually belongs to.

jj-stigell:
closes #320 

Small UI fixes/additions:
* display message if no instances found for the course (prevents loading "limbo" after course creation when no attainments/assessment models exist)
* display loading animation when loading course instances
* display message if no attainments found for the assessment model
* display message if no assessment models found
* add placeholder icons for editing course details (visible only to admin)


client/src/components/CourseView.tsx
```diff
@@ -36,10 +34,8 @@ function CourseView(): JSX.Element {
     useState(null);
   const [assessmentModels, setAssessmentModels]: State<Array<AssessmentModelData>> =
     useState([]);
-  const [attainmentTree, setAttainmentTree]: State<AttainmentData | null> =
-    useState(null);
-
-  const [instances, setInstances]: State<Array<CourseInstanceData>> = useState([]);
+  const [attainmentTree, setAttainmentTree]: State<AttainmentData | null | undefined> =
+    useState(undefined);
```
no-comma, line 38: Why both `null` and `undefined`?

client/src/components/CourseView.tsx
```diff
@@ -55,25 +51,21 @@ function CourseView(): JSX.Element {
...
-        const sortedInstances: Array<CourseInstanceData> = courseInstances.sort(
-          (a: CourseInstanceData, b: CourseInstanceData) => {
-            return sortingServices.sortByDate(a.startDate, b.startDate);
-          }
-        );
-        setInstances(sortedInstances);
+        /**
+         * Newly created courses do not have assessment models assigned.
+         * Set tree to null so proper message can be displayed.
+        */
```
no-comma, line 58:
```suggestion
        /**
         * Newly created courses do not have assessment models assigned.
         * Set tree to null so proper message can be displayed.
         */
```

no-comma (approved):

client/src/components/CourseView.tsx
jj-stigell: My approach was that its undefined when we are fetching the data from API, then we display loading. If we get no assessment models, set it to null to display "no attainments" message, otherwise render attainments.

no-comma: OK, you should add a comment explaining that.

PR #334, opened 2023-07-05
Update course instance creation from client

no-comma:


jj-stigell (approved):
Form looks really nice, I updated the course creation form to use that style. I added warning dialog if cancelling course/instance creation after any of the form fields has been edited.

PR #340, opened 2023-07-06
Remove mock data from select grading formula view

### LINKED ISSUE #246: Front-end should fetch results data from backend

jj-stigell:
* Removes last mockdata used in frontend (select grading view).
* Backend API for getting assessment model attainments updated to have optional "children"/"descendants" param (already exists in the API docs but not implemented in the controller).

closes #246


server/src/controllers/attainment.ts
```diff
@@ -371,6 +372,7 @@ export async function getAttainment(req: Request, res: Response): Promise<void>
 }
 
 export async function getRootAttainment(req: Request, res: Response): Promise<void> {
+  const tree: string = await validateTreeParam(req.query.tree as string);
```
no-comma, line 375: This does no harm again, but I thought the point was to not include this because it seemed unnecessary. At least that's why I removed it initially.

client/src/tests/mock-data/mockAttainments.tsx
no-comma, line 1: With this mock data gone, the remaining attainment data `mockAttainmentsClient.tsx` should be renamed to just `mockAttainments.tsx`.

server/test/controllers/attainment.test.ts
```diff
@@ -899,19 +899,6 @@ describe(
...
-      const res: supertest.Response = await request
-        .get('/v1/courses/2/assessment-models/2/attainments/2?tree=children&foo=bar')
-        .set('Cookie', cookies.userCookie)
-        .set('Accept', 'application/json')
-        .expect(HttpCode.BadRequest);
-      expect(res.body.success).toBe(false);
-      expect(res.body.data).not.toBeDefined();
-      expect(res.body.errors).toBeDefined();
-      expect(res.body.errors[0]).toBe('this field has unspecified keys: foo');
-    });
```
no-comma, line 913: Why is this test removed?

server/src/controllers/attainment.ts
jj-stigell: I needed to get easy access to the root attainment and its immediate children without knowing the actual root attainment id.

server/test/controllers/attainment.test.ts
jj-stigell: I was thinking we can just ignore any extra parameters we do not use.

server/src/controllers/attainment.ts
no-comma: If there turned out to be reason for the feature to exist, then it's fine.

no-comma (approved):

PR #364, opened 2023-07-17
API hooks

no-comma:
- Refactors all API calls to use the [React Query](https://github.com/TanStack/query) library
- Refactors all API services into hooks
- Refactors tests to use the [MSW](https://github.com/mswjs/msw) library for mocking API return values
- Finishes fully porting frontend to backend TypeScript rules


no-comma:
I couldn't get some `FileLoadDialog` tests working, but otherwise everything should be fine.

client/src/components/EditInstanceView.tsx
```diff
@@ -5,70 +5,60 @@
...
 
-  const [instance, setInstance]: State<CourseInstanceData | null> =
-    useState<CourseInstanceData | null>(null);
-
-  const [alertOpen, setAlertOpen]: State<boolean> = useState(false);
-
-  const [messageInfo, setMessageInfo]: State<Message | null> =
-    useState<Message | null>(null);
+  if (!courseId)
+    return (<></>);
```
jj-stigell, line 32: Instead of returning blank pages in missing params, etc. I think we could use the React router [errorElement](https://reactrouter.com/en/main/route/error-element) and create a general use error page.

client/src/tests/mock-data/server.ts
```diff
@@ -0,0 +1,130 @@
...
+import { mockInstances } from './mockInstancesWithStringDates';
+import { mockSisuInstances } from './mockSisuInstances';
+import { mockFormulas } from './mockFormulas';
+
+export function mockSuccess(data: unknown): ResponseResolver<RestRequest, RestContext> {
+  return async (_req: RestRequest, res: ResponseComposition, ctx: RestContext) => {
+    return res(
+      ctx.status(200),
+      ctx.json({
+        success: true,
```
jj-stigell, line 22: Do we even need the `success` attribute in our API?

Quickly checking through the code, only place we use the `success` value is during testing.
Also data and errors properties are never present same time in the backend response, so success is kind of redundant since we can determine the success of the request from properties. Axios doesn't use the `success` but the http code to determine when error must be thrown.

Some API specifications like [JSON api](https://jsonapi.org/format/#document-top-level) even states: "The members `data` and `errors` MUST NOT coexist in the same document." If we use this rule, then there is no need for `success` in my opinion.

client/src/components/course-results-view/SisuExportDialog.tsx
```diff
@@ -94,66 +80,53 @@ export default function SisuExportDialog(props: {
...
-      setSnackPack((prev: Array<Message>) => [...prev, errorMsg]);
-    } finally {
-      setAlertOpen(false);
+  async function handleExportSisuGradeCsv(): Promise<void> {
+    snackPack.push({
+      msg: 'Fetching Sisu CSV...',
+      severity: 'info'
+    });
+
+    exportSisuGradeCsv.refetch();
```
jj-stigell, line 111: invalidating cache. Exporting to Sisu compatible CSV, if used twice, returns the first exports cached file when exporting the second time. This creates problems if in between export calls the list of student numbers, assessmentdate or completionLanguage is changed. We should some how invalidate the cache if any of these values have been modified between exports.

But because this PR is already quite big, the caching strategies should be planned/implemented on some other PR.

jj-stigell (approved):
My comments are out of the scope of this PR or not directly related to it, so I do not require any changes to these but rather have them as memo for future.

client/src/components/EditInstanceView.tsx
no-comma: New issue: #380

client/src/tests/mock-data/server.ts
no-comma: New issue: #381

client/src/components/course-results-view/SisuExportDialog.tsx
no-comma: New issue: #382

PR #369, opened 2023-07-20
Optional instance ID in grade endpoints

### LINKED ISSUE #355: Optional instance ID in grade endpoints
In addition to a list of student numbers, grade endpoints should also allow an optional instance ID query parameter, which will execute the task for the students of that instance.

jj-stigell:
Updates endpoints:
* get sisu csv
* get final grades in JSON
* calculate grades

These endpoints take optional parameter `instanceId`:
* If only instance ID is provided, result are executed for students that are on that particular instance.
* If only student numbers are provided, results are executed for those students, no matter which instance they are on (instance must belong to the assessment model).
* If both instance ID and student number list are provided, results are executed to users belonging to the instance and list of student numbers (intersection).
* If no instance ID or student numbers provided, the all students belonging to any instance that is connected to the assessment model are included in the results.

closes #355


server/src/controllers/grades.ts
```diff
@@ -162,6 +163,47 @@ async function getFinalGradesFor(
...
+async function filterByInstanceAndStudentNumber(
+  instanceId: number,
+  assessmentModelId: number,
+  studentNumbersFiltered: Array<string> | undefined
+): Promise<Array<string> | undefined> {
+  const studentsFromInstance: InstanceWithUsers | null = await CourseInstance.findOne({
+    attributes: ['id'],
+    where: {
+      id: instanceId,
+      assessmentModelId
```
no-comma, line 179: Course instances are not supposed to be strictly tied to assessment models. They only have a reference to a default assessment mode, which likely isn't that important at all from a teacher's point of view, but may be of interest to students later. So this shouldn't require an assessment model ID.

server/src/controllers/grades.ts
```diff
@@ -189,24 +231,38 @@ export async function getSisuFormattedGradingCSV(req: Request, res: Response): P
...
   const [course, assessmentModel]: [Course, AssessmentModel] =
     await validateAssessmentModelPath(
       req.params.courseId, req.params.assessmentModelId
     );
 
   await isTeacherInChargeOrAdmin(req.user as JwtClaims, course.id, HttpCode.Forbidden);
 
-  if (studentNumbers)
-    studentNumbersExist(studentNumbers);
+  // Include students from particular instance (belonging to the assessment model) if ID provided.
```
no-comma, line 257:
```suggestion
  // Include students from a particular instance if an ID is provided.
```

server/src/controllers/grades.ts
jj-stigell: Good point, removed the assessment model id requirement here.

no-comma (approved):

PR #414, opened 2023-08-01
Edit course view

no-comma:
Depends on #405

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -39,7 +39,7 @@
     "yup": "^1.2.0"
   },
   "scripts": {
-    "start": "react-scripts start",
+    "start": "npm --prefix ../common run build && react-scripts start",
     "build": "npm --prefix ../common run build && react-scripts build && node set-web-labels.js",
     "test": "react-scripts test",
     "eject": "react-scripts eject",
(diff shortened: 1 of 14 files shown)


jj-stigell (changes requested):
I noticed that if I try to delete one teacher from the list and nothing else, I get network error. Backend gives me following error `Error: commit has been called on this transaction(ee4ac84d-526b-419c-98cd-6fd102658dda), you can no longer use it. (The rejected query is attached as the 'sql' property of this error)`. 

It seems that transaction is committed before updating the teachers in charge? Does any of the backend tests test just deleting teachers?

no-comma:
> I noticed that if I try to delete one teacher from the list and nothing else, I get network error. Backend gives me following error `Error: commit has been called on this transaction(ee4ac84d-526b-419c-98cd-6fd102658dda), you can no longer use it. (The rejected query is attached as the 'sql' property of this error)`.
> 
> It seems that transaction is committed before updating the teachers in charge? Does any of the backend tests test just deleting teachers?

The error should be fixed now and I added a test case for adding a single teacher in charge and removing a single teacher in charge.

jj-stigell (approved):

PR #460, opened 2023-08-08
Edit grade form

### LINKED ISSUE #451: Edit attainment grades
There are scenarios in which existing grades need to be updated, rather than uploading new ones, for example for correcting exam grading.

There should be an option to edit existing grades. A new API endpoint and a new form in the frontend are needed for this

- [x] #452 
- [ ] #460

jj-stigell:
closes #451


no-comma (changes requested):
I found some bugs when certain grades don't exist leading to runtime errors.

When the final grade for a student doesn't exist, attempting to click edit on it results in an error:
```
props.grade.grades[0] is undefined
EditGradeDialog@http://localhost:3005/static/js/bundle.js:5838:36
...
```

And when a student doesn't have a grade for every non-root attainment, the course results view doesn't load at all and shows an error:
```
grade.grades[0] is undefined
traverseTree@http://localhost:3005/static/js/bundle.js:5254:9
...
CourseResultsTableRow@http://localhost:3005/static/js/bundle.js:5327:54
...
```

Try uploading this CSV to assessment model 1 of course 1 to reproduce the latter error:
```
StudentNumber,name5
999999,5
```

jj-stigell:
First one fixed with conditionally allowing final grade edit.

Second bug seems to exist also in the main branch and is not connected to this feature? Should we open new issue for that?

no-comma:
I don't think a new issue needs to be opened. It should be fixed in #472 already.

no-comma (approved):
I couldn't figure out a way to fix the bug you found, I'll try to solve that in #472 as well.

PR #485, opened 2023-08-15
Attainment grading scale

no-comma:
**Description of changes**
Adds a `maxGrade` column to attainments and moves `minRequiredGrade` from an attainment's `formulaParams` to their own column. In this PR both are stored as integers, though I could see teachers wanting to set `minRequiredGrade` to be a float.

In the CSV upload, a check has been added to ensure that a teacher can't upload a grade which is larger than the `maxGrade` for an attainment, but no such check exists for calculated grades yet (see related issues).

**Requirements**

- [x] I've written tests for new functionality
- [x] I have updated documentation
- [x] My changes to the UI are in accordance with the [design guidelines](
      https://github.com/aalto-grades/base-repository/wiki/Design-Guidelines)
- [x] I have marked all new files with the correct [license](
      https://github.com/aalto-grades/base-repository/wiki/Licensing-Guidelines)

**Related issues**
Closes #299 
Known bug #486

PR #489, opened 2023-08-16
Integer grades

### LINKED ISSUE #459: Final grades should appear as integers
Final grades should be integers (typically from 0 to 5), especially when exported to Sisu as CSV, but also in UI. Now, when pressing "Calculate final grades", the final grades are stored and presented as exact decimal values. They appear as decimal values also in Sisu CSV, but I assume that Sisu would not even accept other than integer values between 0 and 5 (for the majority of courses that use that as the grading scale).

no-comma:
**Description of changes**
Adds a grade type column to attainments which determines whether the grade is supposed to be an integer value or a float value. 

- [x] Use a data type appropriate for both integer and float values in the grade column in attainment grades.
- [x] Sisu CSV files must only include integers.
- [x] The JSON final grade endpoint must only include integers.
- [x] The grade tree endpoint may include either integers or floats for particular attainments, depending on their grade types.
- [x] The CSV upload endpoint should check the grade type of an attainment and validate that the correct grade type was uploaded in the CSV. It should add an error if a float is being uploaded to an attainment which has the integer grade type.
- [x] The grade calculation endpoint should round the calculated grade to the nearest integer in the `calculateFormulaNode` function after calculating a grade using the formula function, if the grade type of the given attainment is an integer.
- [x] The grade editing endpoint should similarly validate the grade type as the CSV upload endpoint. An error should be returned when a float value is being uploaded to an integer grade.

**Requirements**

- [x] I've written tests for new functionality - *REQUIRED FOR THIS PR*
- [x] My changes to the UI are in accordance with the [design guidelines](
- [x] I have marked all new files with the correct [license](
      https://github.com/aalto-grades/base-repository/wiki/Licensing-Guidelines)

**Related issues**
Closes #459 
Depends on #485


no-comma:
I'm not going to be working on this PR anymore. Hopefully the tasks I listed sufficiently describe what I had in mind for this PR.

PR #495, opened 2023-08-16
Multiple sisu export CSV

### LINKED ISSUE #429: Download multiple Sisu CSV files
Related to #427 

When downloading a Sisu grade CSV, if the student list includes students who have already been downloaded in a Sisu CSV previously, those students could instead be included in a second separate CSV file, because teachers are unable to upload grades to Sisu if a grade already exists there. 

What about overriding this in case of mistakes where the grades were never uploaded to Sisu?

jj-stigell:
**Description of changes**
Enable downloading Sisu CSV with only unexported, already exported or all grades selected.

**Requirements**

**Related issues**
Closes #429

### DIFF
--- a/client/src/components/course-results-view/SisuDownloadDialog.tsx
+++ b/client/src/components/course-results-view/SisuDownloadDialog.tsx
@@ -6,7 +6,7 @@ import { FinalGrade, GradeOption, Language } from 'aalto-grades-common/types';
 import {
   Box, Button, Checkbox, Dialog, DialogActions, DialogContent,
   DialogContentText, DialogTitle, FormControlLabel, List, ListItem,
-  ListItemText, MenuItem, Paper, TextField, Typography
+  ListItemText, MenuItem, Paper, TextField, Tooltip, Typography
 } from '@mui/material';
 import { ChangeEvent, JSX, useState } from 'react';
 import { Params, useParams } from 'react-router-dom';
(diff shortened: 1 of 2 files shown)


client/src/components/course-results-view/SisuDownloadDialog.tsx
```diff
@@ -197,20 +214,29 @@ export default function SisuDownloadDialog(props: {
...
-                )} label={(
-                  <Typography variant='body2' sx={{ color: 'red' }}>
-                    The list includes grades already exported to Sisu.
-                    Please check the box if you want to include these previously exported grades.
-                  </Typography>
-                )
-                } />
+              <Box sx={{ ml: 1, my: 2, mr: 10 }}>
+                <Typography variant='body2' sx={{ color: 'red' }}>
+                  The list includes grades already exported to Sisu.
```
no-comma, line 219: I don't expect this to be fixed in this PR, but this should actually be student specific and not grade specific. The described use case for this feature was to cover the following scenario:

Say that student X gets grade 4, and this grade is uploaded to Sisu. Student X wants to raise their grade, and does some extra work to raise their grade to a 5. A new final grade is calculated for student X. Eventually the teacher wants to upload more grades to Sisu, including student X who has already been uploaded before, along with other students who do not have a grade in Sisu yet.

With this current implementation, Grades will not inform the teacher that this student should already have a grade in Sisu. So when the teacher goes to upload this CSV, it will probably fail because teachers themselves are not allowed to raise the grades of students in Sisu and must send off that information for someone else to handle.

no-comma: On further testing this dialog does appear to work correctly on a student basis, but the course results table does not.

no-comma (approved):
I got the impression that this feature should work so that there would be an option to download two CSV files at once, one including unexported students, the other including previously exported students. Either clicking the download button once will download 2 files back-to-back, or maybe it would download a ZIP archive containing two CSV files.

This approach of requiring the teacher to download two files manually may introduce confusion, as the teacher may think that they have to first download the previously exported grades, because if they choose to download unexported grades first, then those grades will also be included in the same category as the grades actually exported previously. Though that doesn't appear to happen with this implementation, I could see that kind of bug appearing in the future.

I'll approve this for now to have something to show rather than nothing, but we should ask about this implementation and potentially change it later.

PR #501, opened 2023-08-17
UI fixes

### LINKED ISSUE #434: Deleting a subattainment incorrectly states "this action cannot be undone"
When deleting a subattainment in the attainment editing view, the deletion dialog states that "this action cannot be undone." When creating an attainment, the message is correct, because the attainment only exists in the frontend's state. However, when editing an attainment, the actual deletion is not done until after "submit" is clicked, meaning the change is not permanent yet. This also does not apply to the root of the tree being deleted, in which case the deletion is done immediately upon clicking "delete."

### LINKED ISSUE #442: Confusing messages about unsaved changes in the course results view
When students have been selected in the course results view, the "return to course view" button turns red and shows a tooltip message:

> You have selected students, these selections will be lost if not saved.

This tooltip doesn't make sense, because the selections can't be saved anyway. This should either be updated to something more appropriate, or it should be evaluated whether the unsaved changes dialog is even needed in the results view in the first place.

### LINKED ISSUE #455: Uploaded grades don't show up until refresh
**To Reproduce**
1. Go to the course results view
2. Successfully upload grades via CSV for a new student

**Expected behavior**
I can immediately see the new student in the table.

**Actual behavior**
The new student is not visible in the table until I refresh the page.

### LINKED ISSUE #467: Global JSX namespace is deprecated
The global `JSX` namespace is deprecated, and `JSX` from `'react'` should be used instead.

jj-stigell:
**Description of changes**
- Fix deprecated JSX
- Show newly uploaded grades after upload in results view
- Update delete/unsaved changes messages. Keeps the unsaved changes dialog in course results view but removes the tooltip.

**Requirements**

**Related issues**
Closes #442 
Closes #434 
Closes #467 
Closes #455

### DIFF
--- a/client/src/components/CourseResultsView.tsx
+++ b/client/src/components/CourseResultsView.tsx
@@ -145,6 +145,7 @@ export default function CourseResultsView(): JSX.Element {
         selectedStudents={selectedStudents}
         setSelectedStudents={setSelectedStudents}
         hasPendingStudents={hasPendingStudents}
+        refetch={students.refetch}
       />
     </Box>
   );
--- a/client/src/components/CourseResultsView.tsx
+++ b/client/src/components/CourseResultsView.tsx
@@ -4,7 +4,7 @@
 
 import { AttainmentData, FinalGrade } from 'aalto-grades-common/types';
 import { Box, Typography } from '@mui/material';
-import { useEffect, useState } from 'react';
+import { JSX, useEffect, useState } from 'react';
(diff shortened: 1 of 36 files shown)


no-comma (approved):
I realized there's another edge case with attainment deletion. When editing an attainment and adding new subattainments, deleting those subattainments is permanent immediately since they haven't been added to the database yet. I updated `cannotBeUndone` in `LeafAttainment` to cover this case.

PR #512, opened 2023-08-18
Creating two assessment models with the same name

### LINKED ISSUE #506: Creating two assessment models with the same name
It is currently possible to create two assessment models for a single course with the same name. This should not be possible, and assessment model names should be unique in each course. It is fine for two assessment models in different courses to have the same name.

jj-stigell:
**Description of changes**
Assessment models have to have unique names on course level.

**Requirements**
- [X] I've written tests for new functionality

**Related issues**
Closes #506


no-comma (changes requested):
I would also include a unique constraint for the assessment model name and course ID.

no-comma (approved):

PR #517, opened 2023-08-21
Csv grade upload dates

### LINKED ISSUE #294: Date entry while uploading grades via CSV
When a teacher is uploading a grade CSV, the frontend should parse attainment tags from the CSV and allow the teacher to set the date and expiry date for each attainment, which are sent in the body to the server.

### LINKED ISSUE #497: Completion date of attainment grades should not be optional
Currently the completion date, or just `date`, of attainment grades is an optional field and can be null.

It should not be optional, as this information is important for all grades.

jj-stigell:
**Description of changes**
Include completion and expiry date entry while uploading grades via CSV. Bulk set for all grades uploaded. Completion date required.

**Requirements**
- [X] I've written tests for new functionality

**Related issues**
Closes #294 
Closes #497


server/src/controllers/grades.ts
```diff
@@ -758,8 +758,21 @@ export function parseGradesFromCsv(
  * the CSV file contains attainments which don't belong to the specified course or course instance.
  */
 export async function addGrades(req: Request, res: Response, next: NextFunction): Promise<void> {
-  /** TODO: Check grading points are not higher than max points of the attainment. */
+  const requestSchema: yup.AnyObjectSchema = yup.object().shape({
+    completionDate: yup
+      .date()
+      .notRequired(),
```
no-comma, line 764: This should not be optional, as it is essential information to know when the grades are no longer valid.

server/src/controllers/grades.ts
```diff
@@ -758,8 +758,21 @@ export function parseGradesFromCsv(
...
+    expiryDate: yup
+      .date()
+      .notRequired()
```
no-comma, line 767: It should be checked that this is greater than or equal to `completionDate`. Now that I think about it, that would be a good constraint to add too. Does Postgres allow for a constraint of the form `Date | null >= Date` (see #497) or `Date | null >= Date | null` (as it is now)?

no-comma (changes requested):
Some notes regarding expiry dates.

With the current implementation of grade expiry, "days valid" of attainments is treated as the primary way to determine whether a grade has expired, meanwhile "expiry date" of attainment grades is a way to override the expiry date given by "days valid."

I think this design places a bit too much attention on setting an expiry date, when it should be optional. Maybe the expiry date box could be hidden by default, requiring the teacher to click something to show it first. This would also be a good place to include the explanation that manually setting an expiry date is for overriding "days valid" of attainments.

I also noticed a bug with uploading grades more than once. If you upload grades and set an expiry date, and then upload grades again without setting an expiry date, instead of the second upload not including an expiry date, the expiry date of the first upload is used.

server/src/controllers/grades.ts
jj-stigell: marked as required, both back and frontend

server/src/controllers/grades.ts
jj-stigell: Added both yup and database checks. Postgres is okay even if there are null values, these are basically just skipped

jj-stigell:
> I think this design places a bit too much attention on setting an expiry date, when it should be optional. Maybe the expiry date box could be hidden by default, requiring the teacher to click something to show it first. This would also be a good place to include the explanation that manually setting an expiry date is for overriding "days valid" of attainments.

By default, expiry date is now hidden.

> I also noticed a bug with uploading grades more than once. If you upload grades and set an expiry date, and then upload grades again without setting an expiry date, instead of the second upload not including an expiry date, the expiry date of the first upload is used.

This bug is fixed.

PR #530, opened 2023-10-12
Use latest attainment grade date for sisu csv

### LINKED ISSUE #424: Use latest attainment date as Sisu CSV date
The option of using the latest attainment date as the date in the Sisu CSV came up. Should this be done or should the teacher still need to set it manually?

kurukimi:
**Description of changes**

**Requirements**

**Related issues**
Closes #424

### DIFF
--- a/server/src/controllers/grades.ts
+++ b/server/src/controllers/grades.ts
@@ -28,7 +28,7 @@ import {
 import { validateAssessmentModelPath } from './utils/assessmentModel';
 import { findAttainmentById, findAttainmentGradeById } from './utils/attainment';
 import { toDateOnlyString } from './utils/date';
-import { gradeIsExpired } from './utils/grades';
+import { gradeIsExpired, getDateOfLatestGrade } from './utils/grades';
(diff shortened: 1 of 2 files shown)


kurukimi:
Now if a date isn't selected while exporting sisu csv, the csv dates become each students latest attainment completion date. The frontend view has the option to select a date, so that that date would be used for all students. We might want to update the frontend view somehow. I thought that the idea to automatically fill the date input with the latest date would not work nicely, since (I think) students can have different latest completion dates of attainments.

PasiSa:
> Now if a date isn't selected while exporting sisu csv, the csv dates become each students latest attainment completion date. The frontend view has the option to select a date, so that that date would be used for all students. We might want to update the frontend view somehow. I thought that the idea to automatically fill the date input with the latest date would not work nicely, since (I think) students can have different latest completion dates of attainments.

Ok, although I think it is quite common that when one uploads a batch of grades to Sisu. They would typically have the same date (e.g. the date of final course exam), but not necessarily always.

PR #545, opened 2023-11-09
Formulas

kurukimi:
**Description of changes**

Add a formula 'rise bonus' along with small fixes and enchancements to the formula making process
**Requirements**

- [ x] I've written tests for new functionality


gubo97000 (approved):
if it passes the dreadful random testes should be good,
worried about the "any" in the types, but if it is unavoidable no probs there :)

PR #620, opened 2024-04-02
Refactor api

bntti:
**Description of changes**
Heavy API refactoring & removal of all unused API endpoints


server/src/types/sisu.ts
gubo97000, line 1: is the entire type even needed

gubo97000 (approved):
lgtm

PR #638, opened 2024-04-18
Edit grades

bntti:
**Description of changes**
Add editing grades dialog


gubo97000 (commented):
why sass? Nested css should be supported in all main browser

Issue #375, opened 2023-07-21
Import points from A+ API


no-comma:
As an A+ teacher, I want to be able to associate a course instance with a course instance in A+, and have an option to fetch partial grades data using A+ API (as an alternative to CSV import). Initially, this could be triggered manually, but later this could happen automatically (e.g., periodically).

PR #661, opened 2024-05-09
Initial A+ integration

no-comma:
**Description of changes**
Adds API endpoints for:
- Fetching exercise data from A+, namely, which modules and difficulties exist in a given A+ course
- Adding A+ grade sources for attainments
- Fetching grades from A+ for a given attainment with an A+ grade source

**Requirements**

- [x] I have marked all new files with the correct [license](
      https://github.com/aalto-grades/base-repository/wiki/Licensing-Guidelines)

**Related issues**
Related to #375


no-comma:
`fetchAplusGrades` in particular could still be improved, but it'd be better to merge this soon and continue work in smaller branches.

server/src/controllers/aplus.ts
```diff
@@ -0,0 +1,217 @@
...
+  } catch (e) {
+    if (e instanceof Error) {
+      throw new ApiError(e.message, HttpCode.BadRequest);
+    }
+    if (e instanceof ZodError) {
+      throw new ApiError(
+        e.issues.map(issue => issue.message),
+        HttpCode.BadRequest
+      );
+    }
```
bntti, line 107: This is never reached because of the above `(e instanceof Error)`.
I would also recommend using zod `safeparse` or `safeparseAsync` to get rid of one try catch.

server/src/controllers/aplus.ts
```diff
@@ -0,0 +1,217 @@
...
+
+    // TODO: There can be multiple sources
+    const gradeSource = await AplusGradeSource.findOne({
+      where: {attainmentId: attainment.id},
+    });
+
+    if (!gradeSource) {
+      throw new ApiError(
+        `attainment with ID ${attainment.id} has no A+ grade sources`,
+        HttpCode.UnprocessableEntity
```
bntti, line 125: Shouldn't this be `404 Not Found`?

common/types/aplus.ts
```diff
@@ -0,0 +1,57 @@
...
+});
+
+export const AplusGradeSourceDataSchema = z
+  .object({
+    attainmentId: z.number().int(),
+    aplusCourseId: z.number().int(),
+    sourceType: AplusGradeSourceTypeSchema,
+    moduleId: z.number().int().optional(),
+    difficulty: z.string().optional(),
+  })
```
bntti, line 32: A [union type](https://zod.dev/?id=discriminated-unions) could be used instead.

server/src/database/migrations/00006-add-aplus-grade-source.ts
```diff
@@ -0,0 +1,72 @@
...
+            primaryKey: true,
+          },
+          attainment_id: {
+            type: DataTypes.INTEGER,
+            allowNull: false,
+            references: {
+              model: 'attainment',
+              key: 'id',
+            },
+            onDelete: 'CASCADE',
```
bntti, line 30: Should probably be `'RESTRICT'` instead

server/src/database/models/aplusGradeSource.ts
```diff
@@ -0,0 +1,76 @@
...
+    updatedAt: DataTypes.DATE,
+  },
+  {
+    sequelize,
+    tableName: 'aplus_grade_source',
+  }
+);
+
+Attainment.hasMany(AplusGradeSource, {
+  onDelete: 'CASCADE',
```
bntti, line 73: Should probably be `'RESTRICT'` instead

server/src/routes/aplus.ts
```diff
@@ -0,0 +1,44 @@
+// SPDX-FileCopyrightText: 2024 The Aalto Grades Developers
+//
+// SPDX-License-Identifier: MIT
+
+import express, {Router} from 'express';
+import {RequestHandler} from 'express-serve-static-core';
```
bntti, line 6: This should be imported from `express`

server/src/routes/aplus.ts
```diff
@@ -0,0 +1,44 @@
...
+  fetchAplusExerciseData,
+  fetchAplusGrades,
+} from '../controllers/aplus';
+import {handleInvalidRequestJson} from '../middleware';
+import {courseAuthorization} from '../middleware/authorization';
+import {controllerDispatcher} from '../middleware/errorHandler';
+
+export const router = Router();
+
+// TODO: Authorization?
```
bntti, line 22: Authorization probably not needed, should handled by the a+ api token.

server/test/controllers/aplus.test.ts
```diff
@@ -0,0 +1,445 @@
...
+
+jest.mock('axios');
+const mockedAxios = axios as jest.Mocked<AxiosStatic>;
+
+beforeAll(async () => {
+  cookies = await getCookies();
+  let _;
+
+  let attainments: AttainmentData[];
+  [courseId, attainments, _] = await createData.createCourse({});
```
bntti, line 47: Here you can just omit the `, _`

server/test/controllers/aplus.test.ts
```diff
@@ -0,0 +1,445 @@
...
+  [fullPointsAttainmentId, moduleAttainmentId, difficultyAttainmentId] =
+    await createData.createAplusGradeSources(courseId);
+  addGradeSourceAttainmentId = attainments[0].id;
+  noGradeSourceAttainmentId = attainments[3].id;
+
+  let otherAttainments: AttainmentData[];
+  [noRoleCourseId, otherAttainments, _] = await createData.createCourse({
```
bntti, line 54: Here you can just omit the `, _`

server/test/controllers/aplus.test.ts
```diff
@@ -0,0 +1,445 @@
...
+      .post(`/v1/courses/${courseId}/aplus-source`)
+      .send([getFullPoints(), getModule(), getDifficulty()])
+      .set('Cookie', cookies.teacherCookie)
+      .expect(HttpCode.Created);
+
+    expect(JSON.stringify(res.body)).toBe('{}');
+    await checkAplusGradeSource(getFullPoints());
+    await checkAplusGradeSource(getModule());
+    await checkAplusGradeSource(getDifficulty());
+  });
```
bntti, line 242: Instead you could just make it a loop `for (const cookie of testCookies)` and remove the duplicate code. There are examples in other test files

server/test/controllers/aplus.test.ts
```diff
@@ -0,0 +1,445 @@
...
+      .get(
+        `/v1/courses/${courseId}/aplus-fetch?attainments=[${fullPointsAttainmentId}]`
+      )
+      .set('Cookie', cookies.teacherCookie)
+      .set('Accept', 'application/json')
+      .expect(HttpCode.Ok);
+
+    const result = await NewGradeArraySchema.safeParseAsync(res.body);
+    expect(result.success).toBeTruthy();
+  });
```
bntti, line 339: Same thing about the cookies here too

server/src/controllers/aplus.ts
no-comma: Good question, I'm not sure. The grade source row isn't exactly a requested resource by the API caller, but rather something that needs to exist for the request to be carried out, so based on how 422 has been used in Grades previously, my first intuition was to use that.

server/src/controllers/aplus.ts
no-comma: The problem here is `JSON.parse`, which will throw an error if `req.query.attainments` is invalid. As far as I know, there is no easy safe way to parse JSON (I'm sure there are libraries though), so we might as well use the same try/catch block for Zod errors as well.

Good catch about this being unreachable. Do we even need to handle the Zod error separately? `ZodError` has a `message` field anyway, which returns `issues` as a JSON string.

bntti (approved):

PR #676, opened 2024-05-17
Add editing final grades

bntti:
**Description of changes**
- Add editing final grades

### DIFF
--- a/client/src/components/course-results-view/GradeCell.tsx
+++ b/client/src/components/course-results-view/GradeCell.tsx
@@ -15,12 +15,10 @@ import {findBestGrade, gradeIsExpired} from '../../utils';
 type GradeCellProps = {
   studentNumber: string;
   attainemntResults?: AttainmentGradesData;
-  finalGrade?: boolean;
 };
 const GradeCell = ({
   studentNumber,
   attainemntResults,
-  finalGrade = false,
 }: GradeCellProps): JSX.Element => {
   const {gradeSelectOption} = useTableContext();
   const theme = useTheme();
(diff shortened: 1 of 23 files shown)


server/docs/finalGrades.yaml
```diff
@@ -131,3 +153,56 @@ definitions:
         $ref: '#/components/responses/Conflict'
     security:
       - cookieAuth: []
+
+/v1/courses/{courseId}/final-grades/{fGradeId}:
+  put:
+    tags: [FinalGrades]
+    description: >
+      Edit a final grade.
+      Available only to admin users and teachers in charge of the course.
```
no-comma, line 162: A general comment, do we need to specify that a feature is accessible to admin users, unless it is exclusively accessible to admin users? Admins should have access to everything anyway, so stating it in every endpoint is a bit redundant.

client/src/components/course-results-view/GradeCell.tsx
```diff
@@ -15,12 +15,10 @@ import {findBestGrade, gradeIsExpired} from '../../utils';
 type GradeCellProps = {
   studentNumber: string;
   attainemntResults?: AttainmentGradesData;
-  finalGrade?: boolean;
 };
 const GradeCell = ({
   studentNumber,
   attainemntResults,
```
no-comma, line 21: Typo

no-comma (commented):
Personally, I would just spell out `finalGradeId` instead of abbreviating it to `fGradeId`. We're already spelling out longer names (`assessmentModelId`) and I want to avoid causing any ambiguity in the future, even if this single abbreviation is fairly innocuous at the moment.

PR #677, opened 2024-05-20
Add renaming assessment models

bntti:
**Description of changes**
- Add support for renaming models

### DIFF
--- a/client/src/components/course-view/EditAssessmentModelDialog.tsx
+++ b/client/src/components/course-view/EditAssessmentModelDialog.tsx
@@ -1,0 +1,100 @@
+// SPDX-FileCopyrightText: 2023 The Aalto Grades Developers
+//
+// SPDX-License-Identifier: MIT
+
+import {
+  Button,
+  CircularProgress,
+  Dialog,
+  DialogActions,
+  DialogContent,
+  DialogTitle,
+  TextField,
+} from '@mui/material';
+import {enqueueSnackbar} from 'notistack';
+import {JSX, useEffect, useState} from 'react';
+import {useParams} from 'react-router-dom';
+
+import {useEditAssessmentModel} from '../../hooks/useApi';
+
+type EditAssessmentModelDialogProps = {
+  assessmentModelId: number | null;
+  name: string | null;
+  onClose: () => void;
+  open: boolean;
+};
+const EditAssessmentModelDialog = ({
+  assessmentModelId,
+  name,
+  onClose,
+  open,
+}: EditAssessmentModelDialogProps): JSX.Element => {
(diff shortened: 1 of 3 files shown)


client/src/components/course-view/EditAssessmentModelDialog.tsx
```diff
@@ -0,0 +1,100 @@
+// SPDX-FileCopyrightText: 2023 The Aalto Grades Developers
```
no-comma, line 1:
```suggestion
// SPDX-FileCopyrightText: 2024 The Aalto Grades Developers
```

client/src/components/course-view/ModelsView.tsx
```diff
@@ -252,19 +263,40 @@ const ModelsView = (): JSX.Element => {
                 secondaryAction={
                   editRights ? (
                     <>
-                      <IconButton
-                        onClick={() =>
-                          handleArchiveModel(model.id, !model.archived)
+                      <Tooltip placement="top" title="Edit assessment model">
```
no-comma, line 266: It becomes fairly obvious once you open the dialog, but should this specify that it's for editing the name specifically? The actual editing is done through the graph, right?

client/src/components/course-view/EditAssessmentModelDialog.tsx
```diff
@@ -0,0 +1,100 @@
...
+import {useParams} from 'react-router-dom';
+
+import {useEditAssessmentModel} from '../../hooks/useApi';
+
+type EditAssessmentModelDialogProps = {
+  assessmentModelId: number | null;
+  name: string | null;
+  onClose: () => void;
+  open: boolean;
+};
```
no-comma, line 25: Have you decided on any consistent style for writing prop types? As in, in some places they are defined as `type PropsType = {...}`, here it is defined as `type ComponentNameProps = {...}`, and in others it's just defined in the component `const ComponentName = (props: {...}) =>`

no-comma (approved):
LGTM, only some minor points and questions.

client/src/components/course-view/EditAssessmentModelDialog.tsx
gubo97000: Until the types are not exported, and only one component exists in a single file PropsType should be good. This case is not always respected though.

client/src/components/course-view/ModelsView.tsx
gubo97000: I agree, "Edit name" or "Rename" would be more clear. If we add more stuff later we can call it edit info.

gubo97000 (approved):

PR #681, opened 2024-05-21
name required in new Attainments modal

gubo97000:


client/src/components/course-view/NewAttainmentDialog.tsx
```diff
@@ -36,10 +37,13 @@ const AddAttainmentDialog = ({
       <DialogTitle>Create new attainment</DialogTitle>
       <DialogContent>
         <TextField
-          sx={{mt: 1}}
+          sx={{mt: 1, '&input:user-invalid': {borderColor: 'red'}}}
```
bntti, line 40: Is this necessary? It seems to work without it

gubo97000: fixed

PR #702, opened 2024-05-24
Prevent invalid final grades

### LINKED ISSUE #692: Don't allow calculating invalid final grades
Calculated final grades should always be valid for the grading scale of the course. E.g., on a pass/fail course, a grade of 2 is invalid, and on a 0-5 scale, a grade of 1.6 is invalid.

bntti:
**Description of changes**
- [x] Fix setting course grading scale
- [x] Prevent adding invalid final grades
- [x] Prevent editing course grading scale if final grades exist

**Related issues**
Closes #692

### DIFF
--- a/server/src/controllers/course.ts
+++ b/server/src/controllers/course.ts
@@ -26,6 +26,7 @@ import {sequelize} from '../database';
 import Course from '../database/models/course';
 import CourseRole from '../database/models/courseRole';
 import CourseTranslation from '../database/models/courseTranslation';
+import FinalGrade from '../database/models/finalGrade';
 import User from '../database/models/user';
 import {ApiError, CourseFull} from '../types';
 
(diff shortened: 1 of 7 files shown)


server/src/controllers/course.ts
```diff
@@ -203,11 +212,12 @@ export const editCourse = async (
...
-          minCredits: minCredits,
-          maxCredits: maxCredits,
-          gradingScale: gradingScale,
-          languageOfInstruction: languageOfInstruction,
+          courseCode: courseCode ?? course.courseCode,
+          minCredits: minCredits ?? course.minCredits,
+          maxCredits: maxCredits ?? course.maxCredits,
+          gradingScale: gradingScale ?? course.gradingScale,
+          languageOfInstruction:
+            languageOfInstruction ?? course.languageOfInstruction,
```
bntti, line 220: This actually does not change the functionality but I think this is more clear? It could also lead to bugs because if one of the values is null it will be ignored (in this case none of them can).

PR #707, opened 2024-05-27
Allow teacher to edit assistants

### LINKED ISSUE #684: Teachers should be able to manage assistants
Currently only admins can add and remove assistants from courses, but this functionality should be available to teachers.

bntti:
**Description of changes**
- Allow teacher to edit assistants

**Related issues**
Closes #684

### DIFF
--- a/client/src/App.tsx
+++ b/client/src/App.tsx
@@ -175,11 +175,7 @@ const router = createBrowserRouter([
           {path: '/', index: true, element: <FrontPage />},
           {
             path: '/:courseId',
-            element: (
-              <>
-                <Outlet />
-              </>
-            ),
+            element: <Outlet />,
             children: [
               {
                 // Temporary default view
@@ -200,11 +196,7 @@ const router = createBrowserRouter([
               },
               {
                 path: '/:courseId/edit',
-                element: (
-                  <PrivateRoute roles={[SystemRole.Admin]}>
-                    <EditCourseView />
-                  </PrivateRoute>
-                ),
+                element: <EditCourseView />,
               },
             ],
           },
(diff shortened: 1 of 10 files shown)


bntti:
Currently a student cannot also be an assistant of a course they have grades in because of the course role unique requirement. I'm pretty sure that is something we need.

no-comma:
> Currently a student cannot also be an assistant of a course they have grades in because of the course role unique requirement. I'm pretty sure that is something we need. It also causes an internal server error, so it should at least be handled better.

This also brings up the question of the access rights of assistants. If assistants have access to all grades of a course, they could go view (or change) their own grades or the grades of their peers back when they were taking the course. This may or may not be desirable.

Last year, we considered the solution that assistants can only view the grades that they themselves have uploaded. Another option is that assistants can only view grades which have been uploaded after they were given the assistant role.

gubo97000:
> > Currently a student cannot also be an assistant of a course they have grades in because of the course role unique requirement. I'm pretty sure that is something we need. It also causes an internal server error, so it should at least be handled better.
> 
> This also brings up the question of the access rights of assistants. If assistants have access to all grades of a course, they could go view (or change) their own grades or the grades of their peers back when they were taking the course. This may or may not be desirable.
> 
> Last year, we considered the solution that assistants can only view the grades that they themselves have uploaded. Another option is that assistants can only view grades which have been uploaded after they were given the assistant role.

There is a code of conduct that employee should respect. I would consider the history system to be the best solution, while giving freedom to the users.

gubo97000 (approved):

gubo97000 (approved):

PR #716, opened 2024-05-30
A+ token

### LINKED ISSUE #672: Store A+ API token from teachers
Where and how (i.e., not in plaintext?) should the token be stored? The user row of the teacher would be the most natural place.

no-comma:
**Description of changes**
Changes the backend A+ endpoints to require the caller to provide an A+ API token and updates the frontend to ask and store the token in local storage.

**Requirements**

- [x] My changes to the UI are in accordance with the [design guidelines](
      https://github.com/aalto-grades/base-repository/wiki/Design-Guidelines) (links open in new tab)

**Related issues**
Fixes #672


server/src/controllers/utils/aplus.ts
```diff
@@ -27,17 +26,41 @@ export const validateAplusCourseId = (aplusCourseId: string): number => {
...
+  const authArray = auth.split(' ');
+  const result = z
+    .tuple([z.literal('Aplus-Token'), z.string().length(40)])
+    .safeParse(authArray);
+  if (!result.success) {
+    throw new ApiError(result.error.message, HttpCode.BadRequest);
+  }
+
+  return authArray[1];
+};
```
no-comma, line 50: Is it fine to use 400 here? As far as Grades is concerned, we only want to know what the token string is, A+ handles all the authentication, so I'm not sure if 401 and 403 are appropriate.

I also need to confirm whether the length of the token is always 40.

bntti: I think 400 is the correct response code here

bntti:
I merged the renaming pr, you should merge dev into this branch and then change `assessment model` to `grading model` and `attainment` to `course part` where applicable

bntti (commented):
Is there any way to clear the API token? What if the teacher typos it somehow?

no-comma:
> Is there any way to clear the API token? What if the teacher typos it somehow?

Clear, no, but you can re-enter it by clicking on your name in the header and then selecting "A+ Token." Should clearing be an option as well?

bntti:
> > Is there any way to clear the API token? What if the teacher typos it somehow?
> 
> Clear, no, but you can re-enter it by clicking on your name in the header and then selecting "A+ Token." Should clearing be an option as well?

If I add an invalid token and then try to re-add a+ course parts, the dialog is just empty and I can't reset it there. I think that should be possible

bntti (approved):

PR #731, opened 2024-06-07
Save Bar for course edit

gubo97000:


client/src/components/shared/SaveBar.tsx
```diff
@@ -0,0 +1,76 @@
...
+  handleDiscard,
+  loading,
+  disabled,
+}: {
+  show: boolean;
+  loading?: boolean;
+  disabled?: boolean;
+  handleSave?: Function;
+  handleDiscard: Function;
+}): JSX.Element => {
```
bntti, line 19: `@deprecated — Use React.JSX instead of the global JSX namespace.`.
I recommend just doing    `import {JSX} from 'react'` instead of using `React.JSX`.

client/src/components/shared/SaveBar.tsx
```diff
@@ -0,0 +1,76 @@
+// SPDX-FileCopyrightText: 2023 The Aalto Grades Developers
+//
+// SPDX-License-Identifier: MIT
+
+import {ClearRounded, SaveOutlined} from '@mui/icons-material';
+import {Box, Button, Fade, Typography, useTheme} from '@mui/material';
+const SaveBar = ({
```
bntti, line 7: Add a newline between the imports and the code (should probably be an eslint rule)

client/src/components/shared/SaveBar.tsx
```diff
@@ -0,0 +1,76 @@
...
+  handleSave,
+  handleDiscard,
+  loading,
+  disabled,
+}: {
+  show: boolean;
+  loading?: boolean;
+  disabled?: boolean;
+  handleSave?: Function;
+  handleDiscard: Function;
```
bntti, line 18: You can use `() => void` instead to specify a better type

client/src/components/shared/SaveBar.tsx
```diff
@@ -0,0 +1,76 @@
...
+        >
+          Save
+        </Button>
+        <Button
+          // variant="tonal"
+          size="small"
+          disabled={loading || disabled}
+          // sx={{float: 'right', mr: 2}}
+          color="error"
+          onClick={() => handleDiscard()}
```
bntti, line 66: `onClick={handleDiscard}`

PR #737, opened 2024-06-11
Warn teacher if editing model with final grades

### LINKED ISSUE #730: Warn teacher about editing a used grading model
After a grading model has been used to calculate grades, it can still be edited. If the model is edited, this may result in different grades being calculated for students who already have grades from before.

The teacher should be shown a warning that they are editing a used grading model and that it may result in different calculations for past grades. The teacher may want this, but if they do not, they should be made aware of it.

bntti:
**Description of changes**
Warn teacher when they try to edit a model with final grades.

**Related issues**
Closes #730

### DIFF
--- a/client/src/components/alerts/SaveConfirmDialog.tsx
+++ b/client/src/components/alerts/SaveConfirmDialog.tsx
@@ -1,0 +1,48 @@
+// SPDX-FileCopyrightText: 2024 The Aalto Grades Developers
+//
+// SPDX-License-Identifier: MIT
+
+import {
+  Button,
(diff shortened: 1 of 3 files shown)


no-comma (approved):
LGMT. You don't need to implement this in this PR, but we may want to show the warning of final grades when the teacher starts editing the model rather than when they're finished editing it and want to save it. Because a teacher may go through a decent amount of effort editing a model only to find out in the end that they don't actually want to change an existing model.

gubo97000 (approved):

PR #739, opened 2024-06-12
Node deletion improvements

### LINKED ISSUE #713: Deleting nodes in the graph is not intuitive
When editing a grading model, it is not obvious how to delete nodes. There doesn't appear to be any indication of how to do it, or if there is, it's easily missable. My first intuition was to use the delete key, but that doesn't do anything.

There should certainly be a button to delete nodes, and I think the delete key should also work as a shortcut.

no-comma:
**Description of changes**
Allow deleting nodes with backspace button and show delete button when nodes are selected

**Related issues**
Fixes #713

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,3 +1,9 @@
+<!--
+SPDX-FileCopyrightText: 2024 The Aalto Grades Developers
+
+SPDX-License-Identifier: MIT
+-->
+
 # Security Policy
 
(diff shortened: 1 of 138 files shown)


gubo97000 (approved):
Some ui changes can be made, like a tooltip on top of the node, but is also true that it doesn't cover the possibility of having more than one node selected

PR #752, opened 2024-06-28
Login-limiter

### LINKED ISSUE #742: Limited number of manual login attempts
The idea is simple the implementation is a bit more complex;

From stackOverflow:

> I like storing failures based on two criteria: if a valid user name was used, count a failure for the user name and IP, otherwise just count a failure for the IP. If a specific account is being attacked then it'll lock out the account independent from the specific IP and any single abusive IP gets a site-wide lockout in the process. It is also important to store an explicit expiration time for locks and renew the lock expiration on each attempt.

Or we use captchas but I have no clue if we can legally implement them.

gubo97000:
Simple limiter of login requests based on IP,
I didn't want to add a user based block to prevent denial of service attack targeted to specific users. 
But we can discuss it, for example we can add a warning log.

Closes #742


bntti:
The backend tests are getting rate limited. You could for example only add the rate limiter if env !== production

bntti (approved):

server/src/middleware/rateLimiterMemory.ts
```diff
@@ -0,0 +1,64 @@
...
+    return originalSend.call(this, body);
+  };
+
+  rateLimiter
+    .consume(req.ip ?? '')
+    .then(() => {
+      next();
+    })
+    .catch(() => {
+      return res.status(429).send({
```
no-comma, line 42: This should be added to the `HttpCode` enum.

PR #771, opened 2024-07-16
One grade per A+ grade source and user

no-comma:
**Description of changes**
Updates the grade creation endpoint to check for existing grades for a student from an A+ grade source, if such grades exist, they will be updated instead of creating new ones.

**Requirements**

**Related issues**
None


no-comma:
Should there be a unique constraint in grades for `userId`, `coursePartId`, and `aplusGradeSourceId`? When `aplusGradeSourceId` is null, there can be multiple grades for the same `userId` and `coursePartId`, does this cause problems?

gubo97000 (approved):

PR #776, opened 2024-07-17
Resolve TODOs

no-comma:
**Description of changes**
Resolves some TODOs, in particular:
- When creating course parts from A+ sources, the state handling between different steps of the process is better
- There is now a default A+ course and instance when asked to select one
- Course part ID is removed from `AplusGradeSourceData` as it's redundant (as opposed to `NewAplusGradeSourceData` where it is necessary)

**Requirements**

**Related issues**
None


client/src/components/course-view/AddAplusGradeSourceDialog.tsx
```diff
@@ -56,13 +56,14 @@ const AddAplusGradeSourceDialog = ({
 
   return (
     <>
-      <Dialog open={open} onClose={handleResetAndClose}>
+      <Dialog open={open} onClose={handleResetAndClose} maxWidth="md">
```
bntti, line 59: I usually also add the `fullWidth` prop but might not be applicable here

client/src/components/course-view/NewAplusCoursePartsDialog.tsx
```diff
@@ -130,20 +129,25 @@ const NewAplusCoursePartsDialog = ({
 
   return (
     <>
-      <Dialog open={open} onClose={handleResetAndClose}>
+      <Dialog open={open} onClose={handleResetAndClose} maxWidth="md">
```
bntti, line 132: I usually also add the `fullWidth` prop but might not be applicable here

gubo97000 (approved):

PR #784, opened 2024-07-22
Add confirm modal

### LINKED ISSUE #636: Create a Global Modal for confirmation of actions
In theory we would need less code than by creating an ad hoc modal every time we will need one.
[react-global-modal](https://github.com/bigyanpoudel/react-global-modal)

bntti:
**Description of changes**
- [x] Add confirm modal
- [x] Use confirm modal to confirm deletes
- [x] Improve unsaved changes modal

**Related issues**
Closes #636


client/src/components/course-view/CoursePartsView.tsx
```diff
@@ -323,6 +331,12 @@ const CoursePartsView = (): JSX.Element => {
       : []),
   ];
 
+  const confirmDiscard = async (): Promise<void> => {
+    if (await AsyncConfirmationModal({confirmNavigate: true})) {
```
gubo97000, line 335: The title and text for the modal is missing.

gubo97000: The text for the modal is missing in many of the discard modals, 
was it intended to have a better default text for the modal than "AsyncConfirmation Modal Title"? Or is it better to add text to every single one?
Otherwise thanks for implementing a better modal system!

bntti: For confirming navigation the body and title can always be the same. I tried to implement a default but forgot to test it. Should be fixed now.

The defaults only trigger when `confirmNavigate=true`, not sure if we want to for example throw an error if the title or message is missing otherwise.

no-comma (approved):

PR #789, opened 2024-07-23
Fetch grades from multiple A+ sources

### LINKED ISSUE #673: Fetching points from multiple A+ grade sources

no-comma:
**Description of changes**
Updates the A+ grade fetching endpoint to fetch grades from all sources of the course parts. In the future, we may want to allow selecting which sources to fetch from, or teachers will have to delete sources that they don't care about anymore to not fetch from them.

**Requirements**

**Related issues**
Fixes #673

### DIFF
--- a/server/src/controllers/aplus.ts
+++ b/server/src/controllers/aplus.ts
@@ -10,6 +10,7 @@ import {TypedRequestBody} from 'zod-express-middleware';
 import {
   AplusCourseData,
   AplusExerciseData,
+  AplusGradeSourceData,
   AplusGradeSourceType,
   HttpCode,
   IdSchema,
@@ -226,7 +227,7 @@ export const fetchAplusGrades = async (
 
     const gradeSources = await AplusGradeSource.findAll({
       where: {coursePartId: coursePart.id},
-    });
+    }) as (AplusGradeSourceData & {date: Date})[];
 
     if (gradeSources.length === 0) {
       throw new ApiError(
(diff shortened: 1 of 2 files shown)


server/src/controllers/aplus.ts
```diff
@@ -223,111 +223,112 @@ export const fetchAplusGrades = async (
       String(coursePartId)
     );
 
-    // TODO: There can be multiple sources
-    const gradeSource = await AplusGradeSource.findOne({
+    const gradeSources = await AplusGradeSource.findAll({
       where: {coursePartId: coursePart.id},
     });
```
bntti, line 228: Add `as AplusGradeSourceData[]` and then you can skip some of the checks below because the cases can never happen

server/src/controllers/aplus.ts
```diff
@@ -223,111 +223,112 @@ export const fetchAplusGrades = async (
...
-              grade = module.points;
-              break;
+      const points = pointsResCache[aplusCourseId];
+      for (const student of points) {
+        // TODO: https://github.com/aalto-grades/base-repository/issues/747
+        if (!student.student_id) {
+          continue;
+        }
+
+        let grade: number | undefined;
```
bntti, line 256: I would prefer the usage of `null` for a value that didn't get a value yet

server/src/controllers/aplus.ts
```diff
@@ -224,109 +225,92 @@ export const fetchAplusGrades = async (
       String(coursePartId)
     );
 
-    // TODO: There can be multiple sources
-    const gradeSource = await AplusGradeSource.findOne({
+    const gradeSources = await AplusGradeSource.findAll({
       where: {coursePartId: coursePart.id},
-    });
+    }) as (AplusGradeSourceData & {date: Date})[];
```
bntti, line 230: Shouldn't date be included in the `AplusGradeSourceData` type?

bntti (approved):

server/src/controllers/aplus.ts
no-comma: If we want to show it to the user when viewing grade sources, then yes. Right now we aren't doing that, though not for any particular reason, it'd be trivial to implement and certainly does no harm.

PR #793, opened 2024-07-24
Max grade

### LINKED ISSUE #779: Re-Introduce maximum grade value for attainments
There are some cases where having a maximum value of the grade would be useful for catching errors. 
For example typos from the teacher or erroneously importing grades of different attainments.

It is not related to the grading model so it cannot be a node; the output would be Error and not Fail.
From A+ we can get the maximum points. So that would not add any overhead in the creation of A+ attainments.
We can also make it optional so teacher are free to choose.

bntti:
**Description of changes**
- [x] Implement max grades for course parts
- [x] Show warnings if grades are higher than max grades
- [x] Improve performance of gradecell
- Does not implement importing max grades from A+

**Related issues**
Closes #779


client/src/components/course-view/UploadDialog.tsx
```diff
@@ -146,27 +167,32 @@ const UploadDialog = ({open, onClose}: PropsType): JSX.Element => {
     onClose();
   };
 
+  console.log(invalidValues);
```
no-comma, line 170: Leftover debug print?

no-comma (approved):

Issue #828, opened 2024-08-14
Course parts


gubo97000:
## Frontend
- [x] #841
- [x] #829
- [x] #831
- [x] #832

## Backend
- [x] #827 
- [x] #838
- [x] #859

## Both
- [x] Add course part models
- [x] #830
- [x] #846

## Common
- [x] #833

## E2E tests
- [x] #839

PR #837, opened 2024-08-15
Course parts

bntti:
**Description of changes**
Tracked in #828


gubo97000:
- When adding multiple course tasks manually they all share the same state so changing the name of one changes the name of the others in the making too

gubo97000 (approved):

PR #851, opened 2024-08-27
Don't log plaintext passwords

no-comma:
**Description of changes**
Omits plaintext passwords from HTTP logs. If the request body is malformed, the password will still get logged, but how big of a concern is that? Technically speaking we can't stop the user from getting their password logged somehow in any case. We could also maybe log the hash of the password rather than omitting it entirely.

**Requirements**
None relevant.

**Related issues**
None.

### DIFF
--- a/server/src/middleware/requestLogger.ts
+++ b/server/src/middleware/requestLogger.ts
@@ -5,6 +5,11 @@
 import {Request, RequestHandler} from 'express';
 import morgan from 'morgan';
 
+import {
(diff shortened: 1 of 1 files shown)


gubo97000:
Yesterday we also talked with Sami about the problem, he suggests to ignore all the request bodies for logs legibility. We should discuss this in more detail.

For now the fix makes sense!
I would wait also for @bntti review

gubo97000 (approved):

PR #852, opened 2024-08-27
Remove client tests

### LINKED ISSUE #813: Remove frontend tests?
Frontend specific tests don't really cover anything which wouldn't also covered by end-to-end tests.

no-comma:
**Description of changes**
Removes all client-specific tests as this is long overdue and if we don't do it soon, it will likely not be completed at all (again) once new developers start working on the system. Also fixes a small line wrapping issue with the PR template.

**Requirements**
None relevant.

**Related issues**
Closes #813


gubo97000:
I would leave the tests there, to be used as guidance for new developers. We can cut them off from the workflow, and write that in the documentation

bntti (approved):

PR #865, opened 2024-09-11
Various small fixes

bntti:
**Description of changes**
- Update passport-saml
- Improve footer links
- Small refactoring
- Show source node value
- Improve graph templates
- Clear course part name when cancelling creation
- Add some grades to the seed data
- Remove extra nodeData when saving graph


bntti:
Some gnu website seems to be down and the build fails because it cannot reach it. The e2e test should pass when run later

gubo97000 (commented):
lgtm

gubo97000 (approved):

PR #892, opened 2024-09-24
Improve workflows

bntti:
#### Description of changes
- [x] Improve github workflows to use [reusable workflows](https://docs.github.com/en/actions/sharing-automations/reusing-workflows)
- [x] Update packages

**Note**
Remember to update the repository settings to point to the new workflows to allow merging pull requests


bntti:
The commit dates are wrong because I had to do a bit of rebasing, but it shouldn't matter since this will be squashed into dev :p

Issue #879, opened 2024-09-20
When downloading Sisu CSV change close button text


bntti:
The button says by default `Cancel` but if downloading CSV such that the dialog doesn't close, the button should say `Close`

PR #900, opened 2024-10-28
Change button text from 'Cancel' to 'Close' conditionally based on selected students

trongduc2206:
If selected student list include exported ones, which means that the dialog won't close after users download CSV file, the text of button is changes from "Cancel" to "Close"

### DIFF
--- a/client/src/components/course/course-results-view/SisuDownloadDialog.tsx
+++ b/client/src/components/course/course-results-view/SisuDownloadDialog.tsx
@@ -249,7 +249,7 @@ const SisuDownloadDialog = ({
       </DialogContent>
       <DialogActions>
         <Button variant="outlined" onClick={onClose}>
-          {t('general.cancel')}
+          {exportedValuesInList ? t('general.close') : t('general.cancel')}
         </Button>
         <Button variant="contained" onClick={handleDownloadSisuGradeCsv}>
           {t('general.download')}


gubo97000 (commented):
When the modal first opens it should always say "Cancel".

Only after the first download it should change to "Close".

Although I understand that it may only add complexity to the code, without altering too much the semantics, so is up to you to decide if the change is good enough.

trongduc2206:
> When the modal first opens it should always say "Cancel".
> 
> Only after the first download it should change to "Close".
> 
> Although I understand that it may only add complexity to the code, without altering too much the semantics, so is up to you to decide if the change is good enough.

I think we are understanding the PR differently. From the description of issue #879, when there are any students that were exported before in the list of selected students to export CSV, there is red text to announce that the dialog won't close. 
image
So I understand that this is the case when the button text should be 'Close' because the dialog will not close. How do you think?

gubo97000:
As I said, I don't want to bike shed on this issue.

But I still think that only after the first press of download in the modal instance the button should change to "close", otherwise there is no meaning in keeping "cancel" in the case with no exported grades.

Example:
button text "cancel" -> *press download* -> now button text is "close"

Otherwise what would be the difference in semantics between closes and cancel on the first opening of the modal?
"Close" should represent a state change in the modal, from a "no action in the current modal has been made" to "an action in the current modal has been made".

Currently when grades have been exported and the modal is opened again (aka when the red text appears) the modal has already the button "close", but it should only have that text after one press of the download button is made, before it should still have the "cancel" text.

But as I said is ok also this way, I prefer to concentrate on more important issues ahah

gubo97000 (approved):

trongduc2206:
> As I said, I don't want to bike shed on this issue.
> 
> But I still think that only after the first press of download in the modal instance the button should change to "close", otherwise there is no meaning in keeping "cancel" in the case with no exported grades.
> 
> Example: button text "cancel" -> _press download_ -> now button text is "close"
> 
> Otherwise what would be the difference in semantics between closes and cancel on the first opening of the modal? "Close" should represent a state change in the modal, from a "no action in the current modal has been made" to "an action in the current modal has been made".
> 
> Currently when grades have been exported and the modal is opened again (aka when the red text appears) the modal has already the button "close", but it should only have that text after one press of the download button is made, before it should still have the "cancel" text.
> 
> But as I said is ok also this way, I prefer to concentrate on more important issues ahah

Alright, I understood your point about the semantics, and yeah as you said, I also think it is better to focus on other important issues

trongduc2206:
Issue #879

PR #940, opened 2025-01-10
enable creating and editing expiryDate for course assistant

trongduc2206:
Expiry date for course assistant can be created or edited

### DIFF
--- a/server/test/controllers/course.test.ts
+++ b/server/test/controllers/course.test.ts
@@ -320,36 +320,25 @@ describe('Test PUT /v1/courses/:courseId - edit course', () => {
     assistants: [],
   };
 
-  const findTeachers = (
+  const findUsers = (
     emails: string[] | undefined
-  ): TeacherData[] | undefined => {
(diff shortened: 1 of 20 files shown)


server/test/controllers/course.test.ts
```diff
@@ -357,8 +375,10 @@ describe('Test PUT /v1/courses/:courseId - edit course', () => {
       ...edit1,
       id: courseId,
       teachersInCharge:
-        findUsers(edit1.teachersInCharge) ?? initState.teachersInCharge,
-      assistants: findUsers(edit1.assistants) ?? initState.assistants,
+        findTeachers(edit1.teachersInCharge) ?? initState.teachersInCharge,
+      assistants:
+        findAssistants(edit1.assistants?.map(assistant => assistant.email)) ??
```
gubo97000, line 380: I have the feeling that only one function can be used, i will check later

trongduc2206: Yes, I read through your commit. It looks great, thank you!

Issue #936, opened 2024-12-17
Rename to Ossi


gubo97000:
Checklist to update the name.
- [ ] URL: aalto-grades.cs.aalto.fi -> ossi.cs.aalto.fi
  - [ ] Update URL for idp
  - [ ] Update VM
  - [ ] Check SSL certificates
- [ ] Refactor Code
- [ ] Documents
- [ ]

jes237:
GitHub wiki page is updated

PR #945, opened 2025-01-21
Non functional Ossi Rename

gubo97000:
Rename in code #936


jes237:
Minor thing I noticed is the favicons, license text itself is updated but the icon design is still "AG".

btw is the wiki also updated to reflect the name change?

gubo97000 (commented):
what is course.spec.t and token.spec.ts? why am i creating them? ahah

jes237:
> @jes237 what is course.spec.t and token.spec.ts? why am i creating them? ahah

spec files should not be there any more, just course.ts and token.ts (reusable test parts)

client/src/components/app-container/Header.tsx
```diff
@@ -53,7 +53,7 @@ const Header = (): JSX.Element => {
         data-testid="a-grades-header-link" // For e2e tests
         viewTransition
       >
-        A{logoVariant} Grades
+        {logoVariant}Ossi
```
jes237, line 56: Is the `logoVariant` needed anymore? It made sense with Aalto logo but not so much with Ossi text.

jes237 (approved):

client/src/components/app-container/Header.tsx
gubo97000: yeah but was fun, in case we remove it later ahah

PR #953, opened 2025-02-10
Fix cell hover

### LINKED ISSUE #938: Grade Cell Re-render on hover
Grade cell uses a state to track hover state of the cursor, this causes re-renders on all the child components, a css solution should be considered

jes237:
#### Description of changes

Use CSS instead of reacts useState to track hovers on iconbutton in the grades table cells.

#### Requirements
- [x] My changes to the UI are in accordance with the [design guidelines](https://github.com/aalto-grades/aalto-grades/wiki/Design-Guidelines)
- [x] I have marked all new files with the correct [license](https://github.com/aalto-grades/aalto-grades/wiki/Licensing-Guidelines)

#### Related issues
Closes #938


gubo97000:
I would suggest to move the icon button with tip to the shared component, folder so we ca use it in other components too. 

and maybe rename it to IconButtonWithTooltip to have a more explicative name

Global classes are ok I guess, if more are needed I would suggest in the future to move them to their own file.

gubo97000 (approved):

